PR #2, opened 2022-12-27
Separate watchdog script from mount helper script

cagrawal-msft:


src/azfswatchdog
```diff
@@ -0,0 +1,160 @@
+#!/bin/bash 
+
+#
```
linuxsmiths, line 3: We have to set some ground rules.
1. The shared data between the mount helper script and the azfs watchdog service will reside in a well known directory inside the tmpfs /run/. We should make a directory named /run/azblobnfs/ and inside that have a file named say, /run/azblobnfs/mountmap.
2. The responsibility of creatin the above directory and the mountmap file lies with the azfs watchdog service, so if the directory and mountmap file doesn't exist the mount helper may bail out with an appropriate error asking the user to check.
3. Since this file would be updated by the service and the mount helper, they should update the file after taking exclusive lock (use the bash command flock for that). Write a function like the following

update_mountmap()
{
    flock /run/azblobnfs/mountmap -c "eval $*";
}

now callers can call it as

update_mountmap sed -i "/ $l_ip $l_nfsip/d" $MOUNTMAP

which will safely update the file
5. chattr will not work for tmpfs so we will have to get rid of that.
6. Once the initial code takes shape, we can think of ways to protect this file from external/undesired mods.
7. All the logs from these files must also go to a log file so that any issues can be debugged. We can start with a log file named /run/azblobnfs/log. In later iterations we will have to think of ways to better rotate this logfile but for now let's keep it simple and start with this single file. We need to write functions for logging, which we can then use in the script instead of the plain echo.
8. f.e. check out the pecho/secho/becho/sbecho/etc variants in 
https://msazure.visualstudio.com/One/_git/Storage-XDataMove-Agent-Linux?path=/packaging/vmrootfs/usr/bin/xdmsh&_a=contents&version=GBmain

src/azfswatchdog
```diff
@@ -0,0 +1,160 @@
...
+
+azfs_watchdog() 
+{ 
+    # Disable signal to make this slightly more resilient. 
+    trap "" INT QUIT TSTP 
+
+    while :; do 
+        sleep $MONITOR_INTERVAL_SECS 
+
+        findmnt=$(findmnt --raw --noheading -o MAJ:MIN,FSTYPE,SOURCE,TARGET,OPTIONS -t nfs)
```
linuxsmiths, line 32: any/all commands used in this file and in the mount helper, should be added as dependency in the package file, so that we are guaranteed that those will exist if our package is installed.

src/azfswatchdog
```diff
@@ -0,0 +1,160 @@
...
+# How often does the watchdog look for unmounts and/or IP address changes for 
+# Blob endpoints. 
+# 
+
+MONITOR_INTERVAL_SECS=5 
+
+# 
+# This stores the map of local IP and share name and external blob endpoint IP. 
+# 
+
```
linuxsmiths, line 13: nit, don't leave this line between comment and the code

src/azfswatchdog
```diff
@@ -0,0 +1,160 @@
...
+# addresses possibly as a result of migration. 
+# 
+# TODO: This must be a separate process started by init/systemd. 
+# 
+
+azfs_watchdog() 
+{ 
+    # Disable signal to make this slightly more resilient. 
+    trap "" INT QUIT TSTP 
+
```
linuxsmiths, line 28: before starting the loop where it monitors MOUNTMAP for changes, it should first create the dir and file if it doesn't exist.

src/mount.azfs
```diff
@@ -266,172 +258,25 @@ update_ipmap()
...
- 
 
 ##### Action starts here ##### 
- 
+
 # [52.239.178.215:/nfsv3stdaccount20/fstestbase3 /mnt/azfs -o rw,tcp,nolock,nconnect=16] 
 echo "Got arguments: [$*]" 
- 
+
 #
```
linuxsmiths, line 269: now we don't need this code that deletes the IPMAP file and tries to kill old watchdog.
It should just check that the watchdog service is running (using "systemctl status azfswatchdog") and if not bail out with an appropriate error, but not try to kill.

src/mount.azfs
```diff
@@ -446,21 +291,23 @@ done
  
 nfs_host=$(get_host_from_share "$1") 
 
- 
+
 if ! is_valid_blob_fqdn $nfs_host; then
```
linuxsmiths, line 295: is_valid_blob_fqdn and many other functions need to be added to this script

linuxsmiths (commented):
I've added only few comment to get started.
Once it starts getting in shape I'll comment on specific things, but first make sure that we have working code. 
Thanks.

src/mount.azfs
```diff
@@ -503,7 +351,7 @@ if [ $? -ne 0 ]; then
     exit 1 
 fi 
 
- 
+
```
linuxsmiths, line 354: don't need to call afs_watchdog now

src/azfswatchdog
cagrawal-msft: Added.

src/mount.azfs
cagrawal-msft: removed.

src/mount.azfs
cagrawal-msft: do you mean those already added functions need to be updated? Since this function is_valid_blob_fqdn is already there.

src/mount.azfs
cagrawal-msft: yes, removed it.

src/azfswatchdog
cagrawal-msft: done.

src/azfswatchdog
cagrawal-msft: Making list of all these dependencies [here](https://microsoft.sharepoint.com/teams/AzureStorage/_layouts/OneNote.aspx?id=%2Fteams%2FAzureStorage%2FShared%20Documents%2FXNfsv3%2FOneNote%2FXNfsv3%20Development&wd=target%28Design.one%7C3751F5D8-2F38-4C12-A46A-62E5EB666E6E%2FDistribution%20of%20Linux%20NFS%20Mount%20Helper%7C4084951E-8752-4491-98BB-BB37FE6601F8%2F%29
onenote:https://microsoft.sharepoint.com/teams/AzureStorage/Shared%20Documents/XNfsv3/OneNote/XNfsv3%20Development/Design.one#Distribution%20of%20Linux%20NFS%20Mount%20Helper&section-id={3751F5D8-2F38-4C12-A46A-62E5EB666E6E}&page-id={4084951E-8752-4491-98BB-BB37FE6601F8}&end). Can start discussing when I start working on those package files.

src/azfswatchdog
cagrawal-msft: updated handling of mountmap in this iteration. Will add proper logging in next one.

cagrawal-msft: Added code to send all the logs into one file which will be in /run/azblobnfs/log as you suggested. To add the shared code between the files in src, I have added "lib" folder in top level which will contain all the helping functions common among the scripts in "src" folder.

src/aznfswatchdog
```diff
@@ -0,0 +1,144 @@
...
+        # If the IP changed for the Blob endpoint, we need to update the DNAT rule.
+        # This will take care of migration/failover causing the Blob endpoint IP to change.
+        #
+        # TODO: Make this less frequent than the unmount check.
+        #
+        # echo "$l_host: new_ip=$new_ip, l_nfsip=$l_nfsip"
+        #
+        if [ "$new_ip" != "$l_nfsip" ]; then 
+            pecho "IP for $l_host changed [$l_nfsip -> $new_ip]."
+            update_mountmap sed -i "/ $l_ip $l_nfsip/d" $MOUNTMAP
```
linuxsmiths, line 129: again
update_mountmap sed -i "/$line/d" $MOUNTMAP

in all places we can delete $line, that's safer

This also allows us to have two methods
add_mountmap
delete_mountmap

which take the entire line as argument and do the >> $MOUNTMAP thing for adding and sed -i thing for deleting. This will not have the issues with sed that you were seeing

also it's better to delete the mountmap line after other things like iptables delete etc, so that in case the program crashes/restarts after that we can correctly take care of the mountmap entry next time around

in all places

src/aznfswatchdog
```diff
@@ -0,0 +1,144 @@
...
+
+            # Delete old DNAT rule and the conntrack entry to stop current active connections too.
+            delete_iptable_entry $l_ip $l_nfsip
+            conntrack -D conntrack -p tcp -d "$l_ip" -r "$l_nfsip"
```
linuxsmiths, line 133: we can move this conntrack inside delete_iptable_entry 
also we can check for the exit status and log accordingly

linuxsmiths (commented):
complete aznfswatchdog
will review mount.aznfs now

src/mount.aznfs
```diff
@@ -0,0 +1,352 @@
...
+    eecho "Cannot resolve IP address for ${nfs_host}!" 
+    exit 1 
+fi
+
+#
+# If user is using private endpoint then we can directly mount without the help
+# of aznfswatchdog service.
+#
+if is_private_ip $nfs_ip; then
+    vecho "Host IP is private ($nfs_ip), using direct NFS mount."
```
linuxsmiths, line 304: ${nfs_host} ($nfs_ip) seems to be using private-link, using direct NFS mount.

src/mount.aznfs
```diff
@@ -0,0 +1,352 @@
...
+fi 
+
+mount_point="$2" 
+mount_options="$4" 
+ 
+get_free_local_ip
+
+if [ $? -ne 0 ]; then
+    eecho "Could not find a free local IP to use for aznfs using DEFAULT_AZNFS_IP_PREFIXES=${DEFAULT_AZNFS_IP_PREFIXES}!"
+    eecho "Set AZNFS_IP_PREFIXES env variable correctly to provide free addresses for use by aznfs!"
```
linuxsmiths, line 324: this log should be different depending on if AZNFS_IP_PREFIXES  env variable is already set.
if not set, the above log is correct, if set, then we should ask user to ensure AZNFS_IP_PREFIXES is set properly

src/mount.aznfs
```diff
@@ -0,0 +1,352 @@
...
+if [ -z "$nfs_dir" ]; then
+    eecho "Bad share name: ${1}."
+    eecho "Share to be mounted must be of the form 'account.blob.core.windows.net:/account/container'." 
+    exit 1 
+fi 
+
+mount_point="$2" 
+mount_options="$4" 
+ 
+get_free_local_ip
```
linuxsmiths, line 320: two parallel mounts might get the same free local ip and then they will try to make conflicting iptable entries.
let's do this
get_free_local_ip must update the mountmap with the correct entry, atomically
so if it returns we know that no other mount could be using the same entry.

To make sure aznfswatchdog doesn't delete the entry after we add the entry and before the mount succeeds, we should add one extra column to this entry in the end. This should contain the pid of the mount process ($$ contains the pid). So aznfswatchdog will check if the entry is of the form with pid and pid exists then it skips the entriy else it deletes the entry.

Then in all failure paths remove that entry and in success path update the entry to remove the extra column.

src/mount.aznfs
```diff
@@ -0,0 +1,352 @@
...
+fi
+
+vecho "nfs_host=[$nfs_host]"
+vecho "nfs_ip=[$nfs_ip]"
+vecho "nfs_dir=[$nfs_dir]"
+vecho "mount_point=[$mount_point]"
+vecho "mount_options=[$mount_options]"
+vecho "local_ip=$LOCAL_IP"
+
+# Check if DNAT rule for forwarding LOCAL_IP traffic to the actual blob endpoint IP already exists.
```
linuxsmiths, line 335: this comment is misleading.. we don't expect it to already exist.
You should say 
Add DNAT rule for forwarding LOCAL_IP traffic to the actual blob endpoint IP address

src/mount.aznfs
```diff
@@ -0,0 +1,352 @@
...
+nfs_dir=$(get_dir_from_share "$1") 
+
+if [ -z "$nfs_dir" ]; then
+    eecho "Bad share name: ${1}."
+    eecho "Share to be mounted must be of the form 'account.blob.core.windows.net:/account/container'." 
+    exit 1 
+fi 
+
+mount_point="$2" 
+mount_options="$4"
```
linuxsmiths, line 318: actually mount options is not always at $4. If user passes option to mount those get shifted, f.e.
try this
mount -w -N 100 -n -v -t aznfs blob.core.windows.net:/acc/con /mnt/aznfs
it logs
Got arguments: [blob.core.windows.net:/acc/con /mnt/aznfs -n -v -o rw -N /proc/1304293/fd/4]

as you can see the -n -v options are added in between.
We will need to scan the arguments for -o and the next one after that is the options.
also the rest of the options we should pass to the mount call so that they are honoured.
you can treat the options as an array of words and then go over them.

following would work

parse_arguments()
{
        # skip share and mountdir.
        shift
        shift
        local next_arg_is_mount_options=false
        OPTIONS=
        MOUNT_OPTIONS=

        for arg in $*; do
                if [ "$arg" == "-o" ]; then
                        next_arg_is_mount_options=true
                        continue
                fi
                if $next_arg_is_mount_options; then
                        MOUNT_OPTIONS=$arg
                        next_arg_is_mount_options=false
                else
                        OPTIONS="$OPTIONS $arg"
                fi
        done
}

then call this at L318 like this

OPTIONS=
MOUNT_OPTIONS=
parse_arguments $*

also need to call the mount command as

`
mount -t nfs $OPTIONS -o "$MOUNT_OPTIONS" "${LOCAL_IP}:${nfs_dir}" "$mount_point" 
`

lib/common.sh
```diff
@@ -0,0 +1,248 @@
...
+{
+    iptables -t nat -D OUTPUT -p tcp -d "$1" -j DNAT --to-destination "$2"
+    if [ $? -ne 0 ]; then 
+        eecho "iptables failed to delete DNAT rule [$1 -> $2]."
+    fi
+}
+
+mkdir -p $RUNDIR
+if [ $? -ne 0 ]; then
+    eecho "Not able to create '${RUNDIR}', Try again."
```
linuxsmiths, line 228: don't say try again.. since we will typically be called from aznfswatchdog for creating these
add [FATAL] in the start
like
eecho "[FATAL] Not able to create ${RUNDIR}."

src/aznfswatchdog
```diff
@@ -0,0 +1,144 @@
...
+    touch $LOGFILE
+fi
+
+# 
+# Watchdog for monitoring unmounts and more importantly change in blob endpoint
+# addresses possibly as a result of migration. 
+#
+while :; do
+    sleep $MONITOR_INTERVAL_SECS
+
```
linuxsmiths, line 28: here we should have a call to
reconcile_mountmap

this function should create a mountmap file from findmnt and iptables o/p..
you can leave that function empty for now and later we will fill it in subsequent revisions

lib/common.sh
```diff
@@ -0,0 +1,248 @@
...
+
+APPNAME="aznfs"
+RUNDIR="/run/${APPNAME}"
+OPTDIR="/opt/microsoft/${APPNAME}"
+LOGFILE="${OPTDIR}/${APPNAME}.log" 
+
+# 
+# This stores the map of local IP and share name and external blob endpoint IP. 
+# 
+MOUNTMAP="${RUNDIR}/mountmap"
```
linuxsmiths, line 11: actually I'm thinking we should go back to storing MOUNTMAP in OPTDIR instead of RUNDIR.
There we have the chattr protection which would avoid most of the common issues with someone accidentally deleting the file.
add_mountmap() and delete_mountmap() should be the only one updating mountmap and so the chattr will be needed only in those files.

src/aznfswatchdog
cagrawal-msft: For now, I have added a TODO.

src/mount.aznfs
cagrawal-msft: I have added functionality for this. Please review it.

src/aznfswatchdog
```diff
@@ -0,0 +1,190 @@
...
+            fi
+            continue
+        fi
+
+        #
+        # This entry was added by get_free_local_ip() call in mount.aznfs to
+        # to ensure two parallel mount requests do not get same local ip. Skip
+        # this entry if PID is active.
+        #
+        if [ ! -z $l_pid ]; then
```
linuxsmiths, line 106: if [ -n "$l_pid" ]; then

use of quotes is very important, else shell will give error

src/aznfswatchdog
```diff
@@ -0,0 +1,190 @@
...
+        #
+        # Delete entry from MOUNTMAP if share is unmounted.
+        # TODO: What if user unmounts and mounts before MONITOR_INTERVAL_SECS secs?
+        #
+        if ! echo "$findmnt" | grep " nfs $m_share " > /dev/null; then
+            # delete the line from MOUNTMAP file.
+            pecho "Deleting unmounted share from MOUNTMAP [$line]."
+            if ! delete_iptable_entry $l_ip $l_nfsip; then
+                eecho "iptables failed to delete DNAT rule [$l_ip -> $l_nfsip]."
+                exit 1
```
linuxsmiths, line 133: we shouldn't exit the service
same for all places where you call exit
it should only exit if it cannot safely continue

In this case it should fall through (after logging) and attempt to delete the mountmap entry.

src/aznfswatchdog
```diff
@@ -0,0 +1,190 @@
...
+        #
+        # echo "$l_host: new_ip=$new_ip, l_nfsip=$l_nfsip"
+        #
+        if [ "$new_ip" != "$l_nfsip" ]; then
+            pecho "IP for $l_host changed [$l_nfsip -> $new_ip]."
+
+            # Delete old DNAT rule and the conntrack entry to stop current active connections too.
+            if ! delete_iptable_entry $l_ip $l_nfsip; then
+                eecho "iptables failed to delete DNAT rule [$l_ip -> $l_nfsip]."
+                exit 1
```
linuxsmiths, line 169: again, no exit

linuxsmiths (commented):
Chandan, I've done a full review. I don't plan to do another full review, so let's sit down today and complete this online.

linuxsmiths (commented):
dummy review

linuxsmiths (approved):

PR #3, opened 2022-12-27
Add Systemd Service file

cagrawal-msft:


src/nfsv3mounthelper.service
```diff
@@ -1 +1,18 @@
...
+Before=
+After=
+
+[Service]
+User=root
+
+#Environment variables
+# [TODO] Does AZFS_IP_PREFIXES var need to be set here? 
+
+ExecStart=usr/sbin/azfswatchdog
```
linuxsmiths, line 14: / missing before usr

PR #5, opened 2023-02-01
Add Packaging code for the Mount helper

cagrawal-msft:


.github/workflows/release.yml
```diff
@@ -0,0 +1,43 @@
...
+    - name: Checkout repository
+      uses: actions/checkout@v3
+    - name: Run Package.sh
+      run: |
+        chmod +x ./package.sh
+        ./package.sh
+      shell: bash
+    - name: Create Release
+      id: crete_release
+      uses: actions/crete_release@v1
```
linuxsmiths, line 25: crete -> create

packaging/aznfs/DEBIAN/control
```diff
@@ -0,0 +1,11 @@
+Package: aznfs
+Version: x.y.z
+Priority: required
+Architecture: amd64
+Origin: Microsoft
+Maintainer: Azure Storage XNFSv3 IDC Team <xnfsv3devsidc@microsoft.com>
+Homepage: https://github.com/Azure/BlobNFS-mount/blob/main/README.md
+Description: Mount helper program to keep NFSv3 mount points working after account migration.
```
linuxsmiths, line 8: this is public facing, so needs to be intentionally subtle, just say this one line.

Mount helper program for correctly handling endpoint IP address changes for Azure Blob NFS mounts.

packaging/aznfs/DEBIAN/postinst
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# Exit on error.
+set -e
+
+echo "===== postinst called. ====="
+
+chmod +x /sbin/aznfswatchdog
+chmod u+s,+x /sbin/mount.aznfs
```
linuxsmiths, line 9: setuid doesn't work for scripts so this not needed.
Did you check if non-superuser is able to use this helper properly when an appropriate entry is added to /etc/fstab?
This is important!!

packaging/aznfs/DEBIAN/postinst
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# Exit on error.
+set -e
+
+echo "===== postinst called. ====="
```
linuxsmiths, line 6: we would want to remove these echos from all these files as they would not look good when user is installing

.github/workflows/release.yml
```diff
@@ -0,0 +1,43 @@
+#
+# Release workflow for Mount Helper Script.
+#
+
```
linuxsmiths, line 4: I can't really review it standalone, since I don't know much about this github pipeline format.
We can  discuss this once you are done with validation

linuxsmiths (commented):
LG mostly. Verification is most important.
Make sure you can install this package correctly on a pristine ubuntu system.

ALSO IMPORTANT - PLS VERIFY non-superuser mount using /etc/fstab works.

.github/workflows/release.yml
```diff
@@ -0,0 +1,44 @@
...
+        prerelease: false
+    - name: Upload Release Asset
+      id: upload-release-asset
+      uses: actions/upload-release-asset@v1
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        upload_url: ${{ steps.crete_release.outputs.upload_url }}
+        asset_path: ./aznfs.deb
+        asset_name: aznfs.deb
```
linuxsmiths, line 44: does it automatically have the release/build number added?
If not, we need to ensure that the artifacts/assets generated are properly tagged

packaging/aznfs/DEBIAN/control
```diff
@@ -0,0 +1,9 @@
+Package: aznfs
+Version: x.y.z
+Priority: required
+Architecture: amd64
+Origin: Microsoft
+Maintainer: Azure Storage XNFSv3 IDC Team <xnfsv3devsidc@microsoft.com>
+Homepage: https://github.com/Azure/BlobNFS-mount/blob/main/README.md
+Description: Mount helper program for correctly handling endpoint IP address changes for Azure Blob NFS mounts.
+Depends: conntrack, iptables, systemd, bind9-host, iproute2, util-linux, nfs-common
```
linuxsmiths, line 9: we cannot put a dependency on systemd since we cannot change the distro's init system if it uses a different one.
what best we can do is check for systemd in preinst checks and fail the install with an appropriate error if someone tries on a non-systemd system.
you can do the following check

init="$(ps -q 1 -o comm=)"
if [ "$init" != "systemd" ]; then
    echo "Cannot install this package on a non-systemd system!"
    exit 1
fi

packaging/aznfs/DEBIAN/postrm
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# Exit on error.
+set -e
+
+chattr -i /opt/microsoft/aznfs/mountmap
+rm -r /opt/microsoft/aznfs
```
linuxsmiths, line 7: as we discussed earlier, we do not control the filesystem of /opt on the user's system, so it may technically not support chattr, so we cannot use "set -e" here, remove "set -e" and check for rm status explicitly.
for chattr add -f to suppress warning 

chattr -if /opt/microsoft/aznfs/mountmap

packaging/aznfs/DEBIAN/prerm
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Exit on error.
+set -e
+
```
linuxsmiths, line 5: we should check for mountmap file and if it has existing mounts, warn the user and ask for confirmation that there are Azure Blob NFS mounts mounted using the aznfs mount helper, they will not be tracked, are you sure you want to continue? stuff like that

.github/workflows/release.yml
cagrawal-msft: I have to test this by running the pipeline in github. that I can only do after this change is merged into default branch of the repo (main branch).

linuxsmiths (approved):

PR #8, opened 2023-02-24
Bug Fix

cagrawal-msft:
Following bugs are being fixed: 
1. Leaving one line after mount.
2. Owner/Group of the files not being root after install
3. Avoid race condition causing entry with PID getting deleted and unmounting the share.
4. Mount taking ~3 secs extra because of ping.

### DIFF
--- a/package.sh
+++ b/package.sh
@@ -36,4 +36,4 @@ cp -avf ${SOURCE_DIR}/src/aznfswatchdog.service ${STG_DIR}/${pkg_dir}${system_di
 cd $STG_DIR
 
 # Create the package.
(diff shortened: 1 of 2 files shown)


linuxsmiths (approved):

linuxsmiths:
> Following bugs are being fixed:
> 
> 1. Leaving one line after mount.
> 2. Owner/Group of the files not being root after install
> 3. Avoid race condition causing entry with PID getting deleted and unmounting the share.
> 4. Mount taking ~3 secs extra because of ping.

3 and 4 are not fixed in this, right?

PR #27, opened 2023-04-05
Handle ZRS SA in case of zone failure

cagrawal-msft:


lib/common.sh
```diff
@@ -152,24 +152,40 @@ resolve_ipv4()
...
+    if [ $cnt_ip -eq 0 ]; then
+        eecho "host returned $cnt_ip address(es) for ${hname}, expected 1 or more!"
+        return 1
     fi
 
-    local ipv4_addr=$(echo "$host_op" | grep " has address " | head -n1 | awk '{print $4}') 
+    local ipv4_addr_all=$(echo "$host_op" | grep " has address " | awk '{print $4}')
+    local ipv4_addr=$(echo "$ipv4_addr_all" | head -n1)
+    
+    if [ $cnt_ip -ne 1 ]
```
linuxsmiths, line 177: if needs a then

lib/common.sh
```diff
@@ -152,24 +152,40 @@ resolve_ipv4()
...
+        for((i=0;i<$cnt_ip;i++))
+        {
+            ipv4_addr=$(echo "$ipv4_addr_all" | head -n1)
+            ping -4 -W3 -c1 $ipv4_addr > /dev/null
+            if [ $? -ne 0 ]; then
+                break
+            fi
+
+            ipv4_addr_all=$(echo "$ipv4_addr_all" | tail -n +$(expr $cnt_ip -1 -$i))
```
linuxsmiths, line 186: instead of updating ipv4_addr_all in every iteration, we can do something like

ipv4_addr=$(echo "$ipv4_addr_all" | tail -n+$i | head -n1)

you will have to now use
for((i=1;i<=$cnt_ip;i++))

also ping cannot be reliably used as a means to find whether node is accessible, since ping doesn't work for blob endpoints, We will have to depend on port 2048 availability.

linuxsmiths (commented):
let's also add SCALABILITY support

lib/common.sh
```diff
@@ -152,24 +152,40 @@ resolve_ipv4()
     local hname="$1"
 
     # Resolve hostname to IPv4 address.
-    host_op=$(host -4 -t A "$hname" | sort) 
+    host_op=$(host -4 -t A "$hname" | sort)
```
linuxsmiths, line 155: for adding SCALABILITY support as discussed in the doc, we might have to do something as simple as

host_op=$(host -4 -t A "$hname" | sort | shuf --random-source=/etc/machine-id)

lib/common.sh
```diff
@@ -152,24 +152,38 @@ resolve_ipv4()
     local hname="$1"
 
     # Resolve hostname to IPv4 address.
-    host_op=$(host -4 -t A "$hname" | sort) 
+    host_op=$(host -4 -t A "$hname" | sort | shuf --random-source=/etc/machine-id)
```
linuxsmiths, line 155: we cannot use /etc/machine-id directly. It may not be present on some machine.
Instead what I'm thinking is we should create a file at install time,

uuidgen > /opt/microsoft/aznfs/machine-id

and then use that file here. That way we are guaranteed that we have a (uniq) file to use for shuffling.
installer can make this file immutable.
uuidgen should mostly be present, if not it can install it first.

also, we can filter out the IP addresses right here and then use $host_op directly in the following code, that way shuffling is more appropriate. So this becomes

host_op=$(host -4 -t A "$hname" | grep " has address " | awk '{print $4}' | sort | shuf --random-source=$OPTDIR/machine-id)

also explain the rationale behind shuffling in the comment.

ALSO NOTE THAT YOU CANNOT DIRECTLY DO THE ABOVE. The host command has to be all by itself, w/o even the sort, o/w we cannot check the return status at L156. So you will have to run the host command by itself and then check the status and if that succeeds run a second command as above, f.e..

host_op=$(host -4 -t A "$hname")

now we can check the return status

now

host_op=$(echo "$host_op" | grep " has address " | awk '{print $4}' | sort | shuf --random-source=$OPTDIR/machine-id)

lib/common.sh
```diff
@@ -152,24 +152,38 @@ resolve_ipv4()
...
 
-    local ipv4_addr=$(echo "$host_op" | grep " has address " | head -n1 | awk '{print $4}') 
+    local ipv4_addr_all=$(echo "$host_op" | grep " has address " | awk '{print $4}')
+    local ipv4_addr=$(echo "$ipv4_addr_all" | head -n1)
+    
+    if [ $cnt_ip -ne 1 ]; then
+        for((i=1;i<=$cnt_ip;i++))
+        {
+            ipv4_addr=$(echo "$ipv4_addr_all" | tail -n +$i | head -n1)
+            nc -w 3 -z $ipv4_addr 2048
```
linuxsmiths, line 181: instead of doing it inline, let's add a function

`#`
`#` Check if a given TCP port is reachable. Uses a 3 secs timeout to bail out in time if address/port is not reachable.
`#`
is_ip_port_reachable()
{
    local ip=$1;
    local port=$2;

    # 3 secs timeout should be good.
    nc -w 3 -z $ip $port
}

lib/common.sh
```diff
@@ -152,24 +152,38 @@ resolve_ipv4()
...
+            if [ $? -ne 0 ]; then
```
linuxsmiths, line 182: you want to break if it's reachable

PR #29, opened 2023-04-08
Major rewrite that now uses single proxy IP per account and not per m…

linuxsmiths:
…ount.

This is very critical change for stability and robustness. It makes many fundamental changes:

1. Use single proxy IP per account irrespective of how many mounts on that account are there from that client, all use the same proxy IP. This makes is analogous to how Linux behaves so many surprises will go away. One of the major reasons for this is to prevent issues with conntrack. If we use one proxy IP per mount then DNAT does not work well since Linux client can use the same port for two different IPs. These clash on the outgoing leg of the DNAT. This cuts down the iptable and mountmap size drastically and helps robustness.
2. Another thing that helps robustness is that we no don't rush to clean up unmounted entries in mountmap and iptables. One of the drivers for that is #1 above since now there are not too many entries in these tables. This simplies the code a lot with all the PID tagged entry logic removed.
3. Now we ensure that mountmap and iptables are always in sync right from the point the mountmap entry is created. When mountmap entry is created iptable entry is also created atomically and when mountmap entry is deleted the iptable entty is also deleted, so at all time they are in sync which simplifies the logic and improves robustness.
4. Many perf improvements by using bash builtins in place of cut/grep/ip tools.
5. resolve_ipv4 is improved for robustness and ZRS support.
6. Logging function simplified.
7. Perf improvement by avoiding portmap calls in the common case.

### DIFF
--- a/src/mountscript.sh
+++ b/src/mountscript.sh
@@ -458,11 +458,9 @@ ensure_aznfswatchdog()
 vecho "Got arguments: [$*]"
 
 # Check if aznfswatchdog service is running.
-vecho "Before ensure_aznfswatchdog"
 if ! ensure_aznfswatchdog; then
     exit 1
 fi
-vecho "After ensure_aznfswatchdog"
 
 # MOUNTMAP file must have been created by aznfswatchdog service.
 if [ ! -f "$MOUNTMAP" ]; then
@@ -484,7 +482,6 @@ if ! is_valid_blob_fqdn "$nfs_host"; then
     exit 1
 fi
 
-vecho "Resolving $nfs_host"
 nfs_ip=$(resolve_ipv4 "$nfs_host")
 if [ $? -ne 0 ]; then
     echo "$nfs_ip"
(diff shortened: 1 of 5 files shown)


src/mountscript.sh
```diff
@@ -275,17 +363,17 @@ search_free_local_ip_with_prefix()
...
-        for ((; _4thoctet<255; _4thoctet++)); do 
-            local_ip="${ip_prefix}.$_4thoctet" 
+        for ((; _4thoctet<255; _4thoctet++)); do
+            local_ip="${ip_prefix}.$_4thoctet"
 
             is_ip_used_by_aznfs=$(echo "$used_local_ips_with_same_prefix" | grep "^${local_ip}$")
             if [ -n "$is_ip_used_by_aznfs" ]; then
-                # Avoid excessive logs. 
+                # Avoid excessive logs.
                 # vecho "$local_ip is in use by aznfs!"
```
cagrawal-msft, line 372: Now we can use log at L372 since duplication of IP is not that frequent now.

cagrawal-msft (approved):

PR #34, opened 2023-04-12
Remove extra code in logging from installer script.

cagrawal-msft:


linuxsmiths (commented):
remove $echoarg from all

linuxsmiths (approved):

PR #51, opened 2023-04-19
Add proper message for user in case of pkg is removed with existing mounts.

cagrawal-msft:
…eted because of existing mounts


packaging/aznfs/RPM/aznfs.spec
```diff
@@ -63,6 +66,11 @@ systemctl daemon-reload
 systemctl enable aznfswatchdog
 systemctl start aznfswatchdog
 
+echo -e "${RED}******************************************************"
+echo -e "Do not uninstall AZNFS in case of existing mounts present." 
+echo -e "It may lead to broken AZNFS package with removed dependencies."
+echo -e "******************************************************${NORMAL}"
+
```
linuxsmiths, line 73: I noticed in one of your screen sharing sessions that the red color is not very readable.
also put an echo before and after this

echo
echo "*******************************************************************"
echo "Do not uninstall AZNFS while you have active aznfs mounts!"
echo "Doing so may lead to broken AZNFS package with unmet dependencies."
echo "If you want to uninstall AZNFS make sure you unmount all aznfs mounts."
echo "********************************************************************"
echo

packaging/aznfs/RPM/aznfs.spec
```diff
@@ -78,6 +86,9 @@ if [ $1 == 0 ]; then
 		echo
 		if [ "$result" != "y" -a "$result" != "Y" ]; then
 			echo "Removal aborted!"
+			echo -e "${RED} Dependencies may have uninstalled required for ANZFS package."
+			echo -e "Please check the required dependencies removed as part of this removal process and install it manually."
+			echo -e "Please contact Microsoft Support if not able to resolve required dependencies.${NORMAL}"
```
linuxsmiths, line 91: echo
echo `"*******************************************************************"`
echo "Unfortunately some of the anzfs dependencies may have been uninstalled."
echo "To fix this you may re-install the latest aznfs package, o/w existing"
echo "aznfs mounts may be affected and new aznfs shares cannot be mounted."
echo `"*******************************************************************"`
echo

package.sh
```diff
@@ -8,6 +8,33 @@
 # Exit on error.
 set -e
 
+postin_message()
+{
+	cat << 'END_HEREDOC'
```
linuxsmiths, line 13: nit, usually EOF is a short-n-sweet delimiter used and no need to put quotes,
also L21 MUST have the delimiter as the only word, 
so

cat << EOF
...
...
EOF

note that the line must start with EOF w/o spaces/tabs before that

linuxsmiths (approved):

linuxsmiths (approved):

PR #59, opened 2023-05-09
Nfsv4 files watchdog

rmirashi-msft:


src/azfilenfs-watchdog
```diff
@@ -0,0 +1,222 @@
...
+# --------------------------------------------------------------------------------------------
+
+#
+# How often does the watchdog look for unmounts and/or to track if stunnel processes are running for
+# nfs file endpoints.
+#
+MONITOR_INTERVAL_SECS=5
+
+#
+# Don't perform mountmap and iptables rule cleanup for unmounted filesystems.
```
hhsubra, line 15: we dont have ip tables

src/azfilenfs-watchdog
```diff
@@ -0,0 +1,222 @@
...
+    IFS=$'\n' lines=$(cat $AZ_FILES_MOUNTMAP)
+    flock -u $fd
+    exec {fd}<&-
+
+    #
+    # findmnt must be done after reading AZ_FILES_MOUNTMAP so that if we come accross a
+    # AZ_FILES_MOUNTMAP entry whose all nfs file shares are unmounted, we know
+    # for sure that it's not in use by any mount and can be removed.
+    #
+    findmnt=$(findmnt | grep 'nfs4\|127.0.0.1' 2>&1)
```
hhsubra, line 122: why hardcode?

src/azfilenfs-watchdog
```diff
@@ -0,0 +1,222 @@
...
+    if [ $? -ne 0 -a -n "$findmnt" ]; then
+        eecho "${findmnt}."
+        eecho "[FATAL] findmnt failed unexpectedly!"
+        eecho "[FATAL] Azfilenfs-watchdog service is exiting, will not monitor Azure NFS file shares."
+        eecho "[FATAL] Please contact Microsoft support before using any NFS File shares."
+        # This usually indicates some non-transient issue, bail out.
+        exit 1
+    fi
+
+    for line in $lines; do
```
hhsubra, line 137: do you need to declare the variable?

src/azfilenfs-watchdog
```diff
@@ -0,0 +1,222 @@
...
+        echo $(stat -c%Y $AZ_FILES_MOUNTMAP)
+    ) 999<$AZ_FILES_MOUNTMAP
+}
+
+
+vecho "Starting azfilenfs-watchdog..."
+
+if [ ! -f $AZ_FILES_MOUNTMAP ]; then
+	touch $AZ_FILES_MOUNTMAP
+	if [ $? -ne 0 ]; then
```
hhsubra, line 83: what if you keep failing with permission issues on creating the mountmap file?

src/azfilenfs-watchdog
```diff
@@ -0,0 +1,222 @@
...
+        # Delete entry from AZ_FILES_MOUNTMAP if there are no mounted shares on that host.
+        # As long as we have at least one mount using the AZ_FILES_MOUNTMAP entry, we leave
+        # it around.
+        #
+        if ! echo "$findmnt" | grep "$accept_port" >/dev/null; then
+                pecho "No mounted shares for host $l_host, deleting from ${AZ_FILES_MOUNTMAP} [$line]."
+
+		# Kill stunnel process first.
+		pid=$(cat $l_pid)
+		kill -9 $pid
```
sharimsft, line 168: script may need privileges to run the findmnt and kill command. how to make sure the scripts have the right privileges to run

src/azfilenfs-watchdog
sharimsft, line 137: lines were not supposed to be greater than certain predefined numbe?

src/azfilenfs-watchdog
```diff
@@ -0,0 +1,222 @@
...
+
+if ! chattr -f +i $AZ_FILES_MOUNTMAP; then
+    wecho "chattr does not work for ${AZ_FILES_MOUNTMAP}!"
+fi
+
+#
+# Watchdog for monitoring unmounts and stunnel processes
+#
+while :; do
+    sleep $MONITOR_INTERVAL_SECS
```
sharimsft, line 97: do we want to check if more than one instance of watchdog script is running so that they dont conflict with each other?

src/azfilenfs-watchdog
```diff
@@ -0,0 +1,222 @@
...
+        else
+	    is_stunnel_running=$(netstat -anp | grep stunnel | grep `cat $l_pid`)
+	    echo "$is_stunnel_running"
+	    if [ -z "$is_stunnel_running" ]; then
+                vecho "stunnel is not running! Restarting the stunnel"
+
+                stunnel_status=$(stunnel $l_conf 2>&1)
+		if [ -n "$stunnel_status" ]; then
+		    eecho "[FATAL] Not able to start stunnel process for '${l_conf}'!"
+                    exit 1
```
sharimsft, line 215: did you try sudo systemctl start stunnel4? instead of having seperate files and start mechanims. all the configuration in one place and just do "sudo systemctl start stunnel4"

rmirashi-msft: If the customer has stunnel.conf files in /etc/stunnel for some other purpose, then running "sudo systemctl start stunnel4" would start all stunnel processes including stunnel processes for /etc/stunnel stunnel.conf files. So I don't think running "sudo systemctl start stunnel4" is a good idea in such scenario.

src/azfilenfs-watchdog
rmirashi-msft: now, we will be having one common watchdog, so there won't be an issue of conflict.

src/azfilenfs-watchdog
rmirashi-msft: We are not enforcing that limit for nfsv4 for now. But we may need to enforce it in future when we will support account migration.

src/azfilenfs-watchdog
rmirashi-msft: We run watchdog script with root privileges, so running kill command wouldn't be a problem. Plus, findmnt can be run without root privileges too.

src/azfilenfs-watchdog
rmirashi-msft: I wanted to add the code related to account migration later, so deleted the code for ip tables from the script. But now, as we will be having common watchdog, the code for ip tables will be resumed.

src/azfilenfs-watchdog
rmirashi-msft: Fixed it in last iteration

src/azfilenfs-watchdog
rmirashi-msft: we run watchdog script with root privileges, so there won't be permission issues on creating the mountmap file. If mountmap file creation fails due to some other reason, we log the error and exit. When customer tries to mount nfsv4 share, mount helper script checks if watchdog is running. If watchdog is not running, mount helper tries to restart watchdog 5 times, and then logs the error and exit if unable to restart the watchdog.

src/aznfswatchdog
```diff
@@ -113,32 +115,96 @@ reconcile_conntrack()
...
+            if [ "$mtime" != "$ifmatch" ]; then
+                eecho "[$1] Refusing to remove from ${AZ_FILES_MOUNTMAP} as $mtime != $ifmatch!"
+                return 1
+            fi
+        fi
 
-if ! chattr -f +i $MOUNTMAP; then
-    wecho "chattr does not work for ${MOUNTMAP}!"
-fi
+	chattr -f -i $AZ_FILES_MOUNTMAP
```
MightyRaymond, line 140: >chattr -f -i $AZ_FILES_MOUNTMAP [](http://example.com/codeflow?start=1&length=31)

why is only this command sticking out?

src/aznfswatchdog
```diff
@@ -330,5 +396,183 @@ while :; do
...
+        IFS=";" read l_host l_conf l_log l_pid l_checksumhash <<< "$line"
+
+        if [ -z "$l_host" -o -z "$l_conf" -o -z "$l_pid" ]; then
+            wecho "[FATAL] Deleting invalid line in $AZ_FILES_MOUNTMAP: [$line]!"
+            l_mtime=$(ensure_az_files_mountmap_not_exist "$line")
+            [ $? -eq 0 ] && mtime_mountmap=$l_mtime
+            continue
+        fi
+
+	accept_port=$(cat $l_conf | grep accept | cut -d ':' -f 2)
```
MightyRaymond, line 462: too many formatting issues, pls fix them first.

MightyRaymond (changes requested):
:clock1:

src/aznfswatchdog
rmirashi-msft: chattr -f -i command is used to remove the immutable attribute from $AZ_FILES_MOUNTMAP file. So whenever we want to modify, delete $AZ_FILES_MOUNTMAP file, we first need to remove the immutable attribute. Then modify or delete $AZ_FILES_MOUNTMAP. And then set the immutable attribute again to prevent the file from being modified or deleted.

src/aznfswatchdog
```diff
@@ -138,45 +140,96 @@ reconcile_conntrack()
...
+#
+# Kill stunnel process and clean up stunnel files generated by aznfs mount helper
+#
+cleanupStunnelFiles()
+{
+    # Kill stunnel process first.
+    pid=$(cat $l_pid)
+    kill -9 $pid
+    if [ $? -ne 0 ]; then
+        eecho "[FATAL] Unable to kill stunnel process '$pid'!"
```
sharimsft, line 206: why this alone in quotes?

src/aznfswatchdog
```diff
@@ -384,5 +437,196 @@ while :; do
...
+            # Update ifmatch time in case of successful updation of AZ_FILES_MOUNTMAP,
+            # so that we can distinguish between AZ_FILES_MOUNTMAP mtime changing because
+            # of our action or some mount helper changing it. In the former case
+            # it's safe to update the AZ_FILES_MOUNTMAP, so update mtime_mountmap to the
+            # mtime after this update.
+            #
+            [ $? -eq 0 ] && mtime_mountmap=$l_mtime
+            continue
+        else
+            # check if checksumHash for stunnel.conf file has changed.
```
sharimsft, line 529: caps C

src/aznfswatchdog
```diff
@@ -384,5 +437,196 @@ while :; do
...
+        eecho "[FATAL] findmnt failed unexpectedly!"
+        eecho "[FATAL] Azfilenfs-watchdog service is exiting, will not monitor Azure NFS file shares."
+        eecho "[FATAL] Please contact Microsoft support before using any NFS File shares."
+        # This usually indicates some non-transient issue, bail out.
+        exit 1
+    fi
+
+    for line in $lines; do
+        if [ -z "$line" ]; then
+            continue
```
zimchen-msft, line 487: where are we remove the empty line?

src/aznfswatchdog
```diff
@@ -384,5 +437,196 @@ while :; do
...
+
+        #
+        # AZ_FILES_MOUNTMAP line is of the form:
+        # account.file.preprod.core.windows.net;<stunnel_account.file.preprod.core.windows.net.conf path>;<stunnel_account.file.preprod.core.windows.net.log path>;<stunnel_account.file.preprod.core.windows.net.pid path>;<checksumHash>
+        #
+        IFS=";" read l_host l_conf l_log l_pid l_checksumhash <<< "$line"
+
+        if [ -z "$l_host" -o -z "$l_conf" -o -z "$l_pid" ]; then
+            wecho "[FATAL] Deleting invalid line in $AZ_FILES_MOUNTMAP: [$line]!"
+            l_mtime=$(ensure_az_files_mountmap_not_exist "$line")
```
zimchen-msft, line 498: Is it intentionally here not passing values for $2? I suggest we always pass value, probably just an empty string or invalid mtime and comment here why that's OK.

src/aznfswatchdog
```diff
@@ -138,45 +140,96 @@ reconcile_conntrack()
...
+ensure_az_files_mountmap_not_exist()
+{
+    (
+        flock -e 999
 
-bash_version=$(bash --version | head -n 1)
-vecho "Linux distribution: $linux_distro"
-vecho "Bash version: $bash_version"
+        #
+        # If user wants to delete the entry only if AZ_FILES_MOUNTMAP has not changed since
```
zimchen-msft, line 153: `I *user* wants`, what is the user here mean? or you mean the watchdog?

src/aznfswatchdog
```diff
@@ -138,45 +140,96 @@ reconcile_conntrack()
...
+            #
+            # If this echo fails then AZ_FILES_MOUNTMAP could be truncated.
+            #
+            echo "$out" > $AZ_FILES_MOUNTMAP
+            ret=$?
+            out=
+            if [ $ret -ne 0 ]; then
+                eecho "*** [FATAL] AZ_FILES_MOUNTMAP may be in inconsistent state, contact Microsoft support ***"
+            fi
+        fi
```
zimchen-msft, line 183: merge with below as `else`?

src/aznfswatchdog
```diff
@@ -138,45 +140,96 @@ reconcile_conntrack()
...
-else
-    linux_distro="Unknown"
-fi
+#
+# Delete entry from $AZ_FILES_MOUNTMAP.
+#
+ensure_az_files_mountmap_not_exist()
+{
+    (
+        flock -e 999
```
zimchen-msft, line 150: what is the purpose of a hardcoded fd?

src/aznfswatchdog
```diff
@@ -384,5 +437,196 @@ while :; do
...
+
+        #
+        # Delete entry from AZ_FILES_MOUNTMAP if there are no mounted shares on that host.
+        # As long as we have at least one mount using the AZ_FILES_MOUNTMAP entry, we leave
+        # it around.
+        #
+        if ! echo "$findmnt" | grep "$accept_port" >/dev/null; then
+            pecho "No mounted shares for host $l_host, deleting from ${AZ_FILES_MOUNTMAP} [$line]."
+
+            cleanupStunnelFiles
```
zimchen-msft, line 514: should we define the `l_host l_conf l_log l_pid l_checksumhash` somewhere that matches the using scope?

src/aznfswatchdog
```diff
@@ -384,5 +437,196 @@ while :; do
...
+                stunnel_status=$(stunnel $l_conf 2>&1)
+                if [ -n "$stunnel_status" ]; then
+                    used_port=$(cat $l_conf | grep accept | cut -d: -f2)
+                    is_binding_error=$(echo $stunnel_status | grep "$LOCALHOST:$used_port: Address already in use")
+                    is_caroot_cert_failure=$(echo $stunnel_status | grep "certificate verify failed")
+                    if [ -n "$is_binding_error" ]; then
+                        eecho "[FATAL] Restarting stunnel failed.$used_port port is already being used by other process.!"
+                        eecho "It's recommended to unmount all shares from $l_host and then remount shares using aznfs mount helper!"
+                    elif [ -n "$is_caroot_cert_failure" ]; then
+                        # TODO: confirm exactly which CA root cert to download for production.
```
zimchen-msft, line 563: this should not be a TODO item for complete this PR

src/aznfswatchdog
sharimsft, line 514: should there be a warning printed for cleanup ? if there was a failure?

sharimsft (approved):

sharimsft (approved):
signoff to merge to feature branch

src/aznfswatchdog
rmirashi-msft: There won't be multiple empty lines. Just one empty line when we unmount all nfsv4 shares (basically when file has no entry after unmounts). Hope that should be okay.

src/aznfswatchdog
rmirashi-msft: Because we are not passing $2 here, that means ensure_az_files_mountmap_not_exist will have $2 = $1, which is a random value (as good as invalid mtime or empty string) and that's okay. Hence not passing $2 intentionally here

src/aznfswatchdog
rmirashi-msft: yes, i mean watchdog

src/aznfswatchdog
rmirashi-msft: hardcoded fd is a random number we chode. There is no purpose as such. I followed v3's style and used hardcoded fd here.

src/aznfswatchdog
rmirashi-msft: scope of l_host l_conf l_log l_pid l_checksumhash are global here. It's not a good practice to keep these variables global. So made it local now.

rmirashi-msft: watchdog calls cleanupStunnelFiles twice.
One is when all the shares from a storageaccount are unmounted. In this scenario, we don't need to warn users about cleanup.
The second scenario is when customer modifies stunnel conf file. In this scenario, we warn customers that we are about to perform cleanup.

src/aznfswatchdog
rmirashi-msft: no, i shouldn't merge with below else. I am getting 'ret' status at two places 172 and 178 line. Hence, if [ $ret -ne 0 ] can be anywhere. Also, the handling of if [ $ret -ne 0 ] is different at both the placess, so I think we shouldn't merge with below else.

src/aznfswatchdog
zimchen-msft: I prefer explicit, but OK to make this as non-blocker review

src/aznfswatchdog
zimchen-msft: won't this line throw client failure? since if you execute this command directly in terminal, there will be failure, can you explain what is the purpose of this line and why this line works?

rmirashi-msft: flock -e is used to acquire exclusive lock on AZ_FILES_MOUNTMAP file. So if you run only flock -e 999, you will get bad descriptor error. You need to redirect 999 to some file. In this code, we are redirecting 999 to AZ_FILES_MOUNTMAP at line 194.
In ensure_az_files_mountmap_not_exist(), we are basically acquiring exclusive lock on AZ_FILES_MOUNTMAP file using flock -e command and doing some execution on AZ_FILES_MOUNTMAP. The lock will be released once the scope or ( ) parentheses ends.
What I meant by there is no purpose as such is there is no purpose to particularly choose 999 file descriptor. You can choose any random number (but in some range) like 123, 999 etc. But you need to redirect that file descriptor to some file else just flock -e <fd> won't succeed.

zimchen-msft (approved):

rmirashi-msft:
> @rmirashi-msft please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@microsoft-github-policy-service agree company="Microsoft"

PR #60, opened 2023-05-23
Mount helper for nfsv4

rmirashi-msft:


src/mountscript.sh
sharimsft, line 1: 20 is the maximum number of lines you can have in the mountmap file

sharimsft, line 1: what is this enable_azure_nconnect going to do?

src/mountscript.sh
```diff
@@ -596,59 +639,429 @@ ensure_aznfswatchdog()
...
+                        eecho "Failed to replace the port in $stunnel_conf_file."
+                        chattr -f +i $stunnel_conf_file
+                        return 1
+                fi
+                chattr -f +i $stunnel_conf_file
+
+                new_used_port=$(cat $stunnel_conf_file | grep accept | cut -d: -f2)
+
+                # start the stunnel process
+		stunnel_status=$(stunnel $stunnel_conf_file 2>&1)
```
sharimsft, line 689: why not just create the file in a location and just use systemctl start stunnel4?

src/mountscript.sh
```diff
@@ -596,59 +639,429 @@ ensure_aznfswatchdog()
...
+
+	stunnel_CAFile="$STUNNELDIR/certs/serverCert.pem"
+        echo "CAFile = $stunnel_CAFile" >> $stunnel_conf_file
+        if [ $? -ne 0 ]; then
+                chattr -f +i $stunnel_conf_file
+                eecho "Failed to add server certificate file path to $stunnel_conf_file!"
+                return 1
+        fi
+
+        echo "verify = 3" >> $stunnel_conf_file
```
sharimsft, line 736: verify option can change depending on user personalization

rmirashi-msft: we are not using verify =3 option. We will be using verifyChain now

src/mountscript.sh
```diff
@@ -596,59 +639,429 @@ ensure_aznfswatchdog()
...
+                        return 1
+                fi
+                chattr -f +i $stunnel_conf_file
+
+                new_used_port=$(cat $stunnel_conf_file | grep accept | cut -d: -f2)
+
+                # start the stunnel process
+		stunnel_status=$(stunnel $stunnel_conf_file 2>&1)
+                if [ -n "$stunnel_status" ]; then
+                        # TODO: Change nfs_host to storageaccount later after server side changes are implemented and merged into Main.
```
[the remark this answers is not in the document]
rmirashi-msft: The mount succeeds only if the stunnel is started/running. First, we start stunnel or check if stunnel is running, and then we mount the share if stunnel is running. If stunnel is not started/running, we log the error and exit, and so we don't mount the share in such case.

src/mountscript.sh
rmirashi-msft: If the customer has stunnel.conf files in /etc/stunnel for some other purpose, and if we run systemctl start stunnel4, that would start all stunnel configurations. So I don't think running systemctl start stunnel4 would be a good solution in such scenario.

src/mountscript.sh
```diff
@@ -217,13 +225,48 @@ get_dir_from_share()
...
+{
+    local mount_options="$MOUNT_OPTIONS"
+    local ver_string="vers="
+    local minor_ver_string="minorversion"
+    local nfs_ver=""
+    local nfs_minorvers=""
+
+    if [ -z "$mount_options" ] || [[ ! "$mount_options" == *"$ver_string"* ]]; then
+        eecho "Bad mount options: ${mount_options}."
+        eecho "Share to be mounted must have version '-o vers=4.1'."
```
MightyRaymond, line 245: >eecho "Share to be mounted must have version '-o vers=4.1'." [](http://example.com/codeflow?start=8&length=60)

can we not default version to 4.1 if tls/notls is specified?

MightyRaymond: chatted with Richa offline, the right thing to do is to default to v3 if vers is not specified.

src/mountscript.sh
```diff
@@ -596,6 +639,324 @@ ensure_aznfswatchdog()
...
+	    break
+        fi
+    done
+}
+
+get_next_available_port()
+{
+    for ((port=NFSV4_PORT_RANGE_START; port<=NFSV4_PORT_RANGE_END; port++))
+    do
+        is_port_available=`netstat -tulpn | grep ":$port "`
```
MightyRaymond, line 694: >l [](http://example.com/codeflow?start=38&length=1)

why only check listening sockets?

src/mountscript.sh
```diff
@@ -596,6 +639,324 @@ ensure_aznfswatchdog()
...
+        # get the next available port
+        available_port=$(get_next_available_port)
+        if [ $? -ne 0 ]; then
+            eecho "Failed to get the next available port for nfsv4 mount."
+            return 1
+        fi
+        vecho "Next Available Port: '$available_port'"
+
+        if [ -z "$available_port" ]; then
+            eecho "Running out of ports. Nfsv4 has port range $NFSV4_PORT_RANGE_START to $NFSV4_PORT_RANGE_END. All ports from this range are used by other processes."
```
MightyRaymond, line 655: >Nfsv4 has [](http://example.com/codeflow?start=41&length=9)

pls change it to "stunnel uses" or something.

src/mountscript.sh
```diff
@@ -596,6 +639,324 @@ ensure_aznfswatchdog()
...
+        stunnel_status=$(stunnel $stunnel_conf_file 2>&1)
+        if [ -n "$stunnel_status" ]; then
+            # TODO: Change nfs_host to storageaccount later after server side changes are implemented and merged into Main.
+            is_binding_error=$(echo $stunnel_status | grep "$LOCALHOST:$new_used_port: Address already in use")
+            if [ -z "$is_binding_error" ]; then
+                eecho "[FATAL] Not able to start stunnel process for '${stunnel_conf_file}'!"
+                return 1
+            fi
+        else
+	    vecho "Found new port '$new_used_port' and restarted stunnel."
```
MightyRaymond, line 684: >vecho "Found new port '$new_used_port' and restarted stunnel." [](http://example.com/codeflow?start=5&length=62)

indent

src/mountscript.sh
```diff
@@ -596,6 +639,324 @@ ensure_aznfswatchdog()
...
+                find_next_available_port_and_start_stunnel "$stunnel_conf_file"
+		is_stunnel_running=$?
+		if [ $is_stunnel_running -ne 0 ]; then
+		    eecho "Failed to get the next available port and start stunnel."
+		    chattr -i -f $stunnel_conf_file
+		    rm $stunnel_conf_file
+		    exit 1
+		fi
+	    fi
+	fi
```
MightyRaymond, line 887: the alignment of if..else seems messed up, pls reformat

src/mountscript.sh
```diff
@@ -596,6 +639,324 @@ ensure_aznfswatchdog()
...
+            else
+                find_next_available_port_and_start_stunnel "$stunnel_conf_file"
+		is_stunnel_running=$?
+		if [ $is_stunnel_running -ne 0 ]; then
+		    eecho "Failed to get the next available port and start stunnel."
+		    chattr -i -f $stunnel_conf_file
+		    rm $stunnel_conf_file
+		    exit 1
+		fi
+	    fi
```
MightyRaymond, line 886: here too, hard to understand the which "if" block they are for.

src/mountscript.sh
```diff
@@ -596,6 +639,324 @@ ensure_aznfswatchdog()
...
+	        else
+		    find_next_available_port_and_start_stunnel "$stunnel_conf_file"
+		    is_stunnel_running=$?
+		    if [ $is_stunnel_running -ne 0 ]; then
+		        eecho "Failed to get the next available port and start stunnel."
+		        exit 1
+		    fi
+	        fi
+	    fi
+        fi
```
MightyRaymond, line 940: please fix alignment/formatting

src/mountscript.sh
```diff
@@ -596,6 +639,324 @@ ensure_aznfswatchdog()
...
+        available_port=$(cat $stunnel_conf_file | grep accept | cut -d: -f2)
+        vecho "Local Port to use: $available_port"
+    fi
+
+    mount_output=$(mount -t nfs -o "$MOUNT_OPTIONS,port=$available_port" "${LOCALHOST}:${nfs_dir}" "$mount_point" 2>&1)
+    mount_status=$?
+
+    if [ -n "$mount_output" ]; then
+        pecho "$mount_output"
+	vecho "Mount completed: ${LOCALHOST}:${nfs_dir} on $mount_point with port:${available_port}"
```
MightyRaymond, line 951: >vecho [](http://example.com/codeflow?start=1&length=5)

missing indent? please review the format, see too many issues like this

MightyRaymond (changes requested):
:clock1:

lib/common.sh
```diff
@@ -15,6 +15,11 @@ RANDBYTES="${OPTDIR}/randbytes"
 #
 MOUNTMAP="${OPTDIR}/mountmap"
 
+#
+# This stores the map of hostname and stunnel conf, log, pid files paths.
+#
+AZ_FILES_MOUNTMAP="$OPTDIR/aznfs_files_mountmap"
```
MightyRaymond, line 21: >AZ_FILES_MOUNTMAP [](http://example.com/codeflow?start=0&length=17)

is this being used?

rmirashi-msft: yes, its being used in mountscript.sh

PR #75, opened 2023-07-27
changes for installation inside container. Allow installation if systemd is installed but may not be booted up.

palashvij-msft:


packaging/aznfs/DEBIAN/preinst
cagrawal-msft, line 1: This change is already there in Tomar's PR. Not needed again.

packaging/aznfs/DEBIAN/postinst
```diff
@@ -35,6 +35,22 @@ chattr +i /opt/microsoft/aznfs/randbytes
...
+    # Reload systemd configuration to ensure any changes take effect.
+    systemctl daemon-reload
+    
+    # Enable the 'aznfswatchdog' service to start on boot.
+    systemctl enable aznfswatchdog
+    
+    # Start the 'aznfswatchdog' service.
+    systemctl start aznfswatchdog
+else
+    echo "OK, systemd installed, but not booted, continuing installation!"
```
cagrawal-msft, line 52: This same log line will come twice in the same panel if kept here as well since it is already there in the end of preinst script. Remove it.

packaging/aznfs/DEBIAN/postinst
```diff
@@ -35,6 +35,22 @@ chattr +i /opt/microsoft/aznfs/randbytes
...
 
 # Start aznfswatchdog service.
-systemctl daemon-reload
-systemctl enable aznfswatchdog
-systemctl start aznfswatchdog
\ No newline at end of file
+init="$(ps -q 1 -o comm=)"
+
+if [ "$init" == "systemd" ]; then
+    echo "System is booted with systemd, continuing installation!"
```
cagrawal-msft, line 41: I don't think this is also necessary. In preinst script, we already have log mentioning system is not booted with systemd.

linuxsmiths: yes, don't repeat the log. I'd suggest the following, with comments and appropriate log

`
#
# If we are not booted with systemd that most likely means we are in a
# container that's being created. Avoid doing any changes that need
# systemd to be running, just enable the aznfswatchdog service so that
# it gets started when the container is booted with systemd.
#
if [ "$init" == "systemd" ]; then
    # Reload systemd configuration to ensure any changes take effect.
    systemctl daemon-reload
    
    # Enable the 'aznfswatchdog' service to start on boot.
    systemctl enable aznfswatchdog
    
    # Start the 'aznfswatchdog' service.
    systemctl start aznfswatchdog
else
    # Enable the 'aznfswatchdog' service to start on boot (not starting it since init is not systemd).
    systemctl enable aznfswatchdog

    echo "aznfswatchdog enabled but not started, will start when system is booted with systemd init!"    
fi
`

PR #77, opened 2023-08-09
Adding support for arm64 architecture for AKS

palashvij-msft:


linuxsmiths (approved):

package.sh
```diff
@@ -66,11 +66,56 @@ generate_rpm_package()
...
+
+    if [ "$arch" == "amd64" ]; then
+        tar_pkg_dir=$tar_pkg_dir_amd64
+        compiler="gcc"
+    elif [ "$arch" == "arm64" ]; then
+        tar_pkg_dir=$tar_pkg_dir_arm64
+        compiler="aarch64-linux-gnu-gcc"
+    else
+        echo "Unsupported architecture: $arch"
+        exit 1
```
linuxsmiths, line 82: nit, just to be consistent with other places, just fail this function instead of exiting.
so, just say

return 1

package.sh
```diff
@@ -66,11 +66,56 @@ generate_rpm_package()
...
+    tar -czvf ${STG_DIR}/tarball/${tar_pkg_dir}.tar.gz *
+}
+
 #STG_DIR, RELEASE_NUMBER and SOURCE_DIR will be taken as env var.
 pkg_name="aznfs"
 pkg_dir="${pkg_name}-${RELEASE_NUMBER}-1_amd64"
 rpm_pkg_dir="${pkg_name}-${RELEASE_NUMBER}-1.x86_64"
-tar_pkg_dir="${pkg_name}-${RELEASE_NUMBER}-1.x86_64"
+tar_pkg_dir_amd64="${pkg_name}-${RELEASE_NUMBER}-1.x86_64"
+tar_pkg_dir_arm64="${pkg_name}-${RELEASE_NUMBER}-1.arm64"
```
cagrawal-msft, line 118: No need to create two separate var names here. In generate_tarball_package(), you can assigne the correct values to "tar_pkg_dir" variable.

cagrawal-msft: Is it visible now?

cagrawal-msft (approved):

linuxsmiths (approved):

PR #78, opened 2023-08-24
Azure/personal/dns warning

palashvij-msft:


src/mountscript.sh
```diff
@@ -636,6 +636,16 @@ if ! is_valid_blob_fqdn "$nfs_host"; then
...
 
+#
+# Fail mount before resolution if the hostname entry is present in /etc/hosts
+#
+if is_present_in_etc_hosts "$nfs_host"; then
+    eecho "Mount failed!"
+    eecho "Detected entry for $nfs_host in /etc/hosts."
+    eecho "[Action Required]: Remove or comment out the entry for $nfs_host in /etc/hosts for MOUNT to work."
+    exit 1
+fi
```
linuxsmiths, line 647: we need to be very careful that we don't raise a red flag, so let's first resolve the IP and then check for the entire
IP[[:space:]]+fqdn so that we only complain when user has set the resolver config such that the address is resolving from /etc/hosts.

lib/common.sh
```diff
@@ -134,6 +134,26 @@ is_ip_port_reachable()
...
+#
+is_present_in_etc_hosts() {
+    local ip="$1"
+    local hostname="$2"
+
+    while IFS= read -r entry; do
+        # Check if the line is not commented
+        if ! [[ "$entry" =~ ^[[:space:]]*# ]]; then
+            # Check if corresponding entry for IP and hostname is present
+            if echo "$entry" | grep -qE "^[[:space:]]*$ip[[:space:]]*$hostname"; then
```
cagrawal-msft, line 148: @linuxsmiths should the regex be "^[[:space:]]*$ip[[:space:]]*$hostname**$**" to avoid anything else after this host name?

cagrawal-msft: A valid format for /etc/hosts could be "_ip-address FQDN Placeholder-for-FQDN_"
@linuxsmiths we have to handle this case as well right?

src/mountscript.sh
```diff
@@ -636,9 +636,11 @@ if ! is_valid_blob_fqdn "$nfs_host"; then
     exit 1
 fi
 
-nfs_ip=$(resolve_ipv4 "$nfs_host")
+# Resolve the IP address for the NFS host
+nfs_ip=$(resolve_ipv4 "$nfs_host" "true")
 if [ $? -ne 0 ]; then
     echo "$nfs_ip"
+    eecho "Mount failed!"
```
cagrawal-msft, line 643: This should be the last comment in case mount fails. Add it after L644 here.

lib/common.sh
```diff
@@ -134,6 +134,21 @@ is_ip_port_reachable()
...
 }
 
+#
+# Verify if FQDN is resolved into ipv4_addr by /etc/hosts entry.
+#
+is_present_in_etc_hosts() {
+    local ip="$1"
+    local hostname="$2"
+
+    # search for the entry in /etc/hosts
```
cagrawal-msft, line 144: nit: Search

lib/common.sh
```diff
@@ -213,6 +229,21 @@ resolve_ipv4()
...
+    # 
+    if is_present_in_etc_hosts "$ipv4_addr" "$hname"; then
+        if [ "$fail_if_present_in_etc_hosts" == "true" ]; then
+            eecho "[ERROR] Detected entry $ipv4_addr $hname in /etc/hosts."
+            eecho "$hname is resolving into $ipv4_addr from /etc/hosts."
+            eecho "[Action Required]: Remove the entry for "$hname" in /etc/hosts to ensure proper handling of endpoint IP address changes by AZNFS."
+            return 1
+        else
+            wecho "[WARNING] Detected entry $ipv4_addr $hname in /etc/hosts." 1>/dev/null
+            wecho "[Action Required]: Remove the entry for "$hname" in /etc/hosts to ensure proper handling of endpoint IP address changes by AZNFS." 1>/dev/null
```
cagrawal-msft, line 243: We have not used [Error] and other such things {except FATAL} in our code anywhere. Lets be consistent. I already have a workitem for such improvements in log. Lets do it for the whole code separately if we get time later.

palashvij-msft: yes, makes sense.

lib/common.sh
```diff
@@ -134,6 +134,21 @@ is_ip_port_reachable()
...
+is_present_in_etc_hosts() {
+    local ip="$1"
+    local hostname="$2"
+
+    # Search for the entry in /etc/hosts
+    if grep -qE "^[[:space:]]*$ip[[:space:]]+[^#]*\<$hostname\>" /etc/hosts; then
+        return 0
+    fi
+
+    return 1
```
linuxsmiths, line 149: nit,
you can simply replace this with
grep -qE "^[[:space:]]*$ip[[:space:]]+[^#]*\<$hostname\>" /etc/hosts

lib/common.sh
```diff
@@ -134,6 +134,21 @@ is_ip_port_reachable()
     nc -w 3 -z $ip $port > /dev/null 2>&1
 }
 
+#
+# Verify if FQDN is resolved into ipv4_addr by /etc/hosts entry.
+#
+is_present_in_etc_hosts() {
```
linuxsmiths, line 140: nit, move the brace to the next line for consistency with other places

lib/common.sh
```diff
@@ -134,6 +134,21 @@ is_ip_port_reachable()
     nc -w 3 -z $ip $port > /dev/null 2>&1
 }
 
+#
+# Verify if FQDN is resolved into ipv4_addr by /etc/hosts entry.
```
linuxsmiths, line 138: ipv4_addr -> IPv4 address

lib/common.sh
```diff
@@ -213,6 +229,21 @@ resolve_ipv4()
...
+    # 
+    if is_present_in_etc_hosts "$ipv4_addr" "$hname"; then
+        if [ "$fail_if_present_in_etc_hosts" == "true" ]; then
+            eecho "[FATAL] Detected entry $ipv4_addr $hname in /etc/hosts."
+            eecho "$hname is resolving into $ipv4_addr from /etc/hosts."
+            eecho "Remove the entry for "$hname" in /etc/hosts to ensure proper handling of endpoint IP address changes by AZNFS."
+            return 1
+        else
+            wecho "[FATAL] Detected entry $ipv4_addr $hname in /etc/hosts." 1>/dev/null
+            wecho "Remove the entry for "$hname" in /etc/hosts to ensure proper handling of endpoint IP address changes by AZNFS." 1>/dev/null
```
linuxsmiths, line 243: remove dot at the end of each log, we don't add dot at the end of logs in other places, isn't it?
also why do you want to suppress log coming on stdout at L242 243, let it come on stdout

lib/common.sh
```diff
@@ -213,6 +229,21 @@ resolve_ipv4()
...
     fi
 
+    #
+    # Check if the IP-FQDN pair is present in /etc/hosts
+    # 
+    if is_present_in_etc_hosts "$ipv4_addr" "$hname"; then
+        if [ "$fail_if_present_in_etc_hosts" == "true" ]; then
+            eecho "[FATAL] Detected entry $ipv4_addr $hname in /etc/hosts."
+            eecho "$hname is resolving into $ipv4_addr from /etc/hosts."
+            eecho "Remove the entry for "$hname" in /etc/hosts to ensure proper handling of endpoint IP address changes by AZNFS."
```
linuxsmiths, line 239: eecho "[FATAL] $hname resolved to $ipv4_addr from /etc/hosts!"
 eecho "AZNFS depends on dynamically detecting DNS changes for proper handling of endpoint address changes"
 eecho "Please remove the entry for $hname from /etc/hosts"

lib/common.sh
palashvij-msft: Here, I have suppressed the this log, as otherwise value of new_ip in aznfswatchdog gets updated to this error log
L 315:  new_ip=$(resolve_ipv4 "$l_host" "false").

linuxsmiths (approved):

PR #79, opened 2023-08-24
Adding double quotes

palashvij-msft:
Bug on line 140 in aznfswatchdog script: [: too many arguments
_if [ $? -eq 0 -a $conntrack_timeo_timew -gt $AZNFS_TIMEWAIT_TIMEOUT ]; then_

Solution:
Wrap the variable in double quotes, forcing it to stay as one string (therefore one argument).

### DIFF
--- a/src/aznfswatchdog
+++ b/src/aznfswatchdog
@@ -129,6 +129,8 @@ reconcile_conntrack()
     done
 }
 
+vecho "Starting aznfswatchdog..."
+
 # Detect and log distro and bash version
 if [ -f /etc/centos-release ]; then
     linux_distro=$(cat /etc/centos-release)
@@ -141,8 +143,6 @@ fi
 bash_version=$(bash --version | head -n 1)
 vecho "Linux distribution: $linux_distro"
 vecho "Bash version: $bash_version"
-
-vecho "Starting aznfswatchdog..."
 
 # Dump NAT table once on startup in case we have reported conflicts.
 vecho "NAT table:\n$(iptables-save -t nat)"
--- a/src/aznfswatchdog
+++ b/src/aznfswatchdog
@@ -137,7 +137,7 @@ conntrack -L > /dev/null
 
 # conntrack timewait timeout higher than the TCP timewait timeout value isn't very valuable.
 conntrack_timeo_timew=$(cat /proc/sys/net/netfilter/nf_conntrack_tcp_timeout_time_wait 2>/dev/null)
-if [ $? -eq 0 ] && [ "$conntrack_timeo_timew" -gt "$AZNFS_TIMEWAIT_TIMEOUT" ]; then
+if [ $? -eq 0 ] && [ $conntrack_timeo_timew -gt $AZNFS_TIMEWAIT_TIMEOUT ]; then
         vecho "Changing /proc/sys/net/netfilter/nf_conntrack_tcp_timeout_time_wait [$conntrack_timeo_timew -> $AZNFS_TIMEWAIT_TIMEOUT]"
         echo $AZNFS_TIMEWAIT_TIMEOUT > /proc/sys/net/netfilter/nf_conntrack_tcp_timeout_time_wait
 fi
--- a/src/aznfswatchdog
+++ b/src/aznfswatchdog
@@ -130,7 +130,9 @@ reconcile_conntrack()
 }
 
 # Detect and log distro and bash version
-if [ -r /etc/os-release ]; then
+if [ -f /etc/centos-release ]; then
+    linux_distro=$(cat /etc/centos-release)
+elif [ -f /etc/os-release ]; then
     linux_distro=$(grep "^PRETTY_NAME=" /etc/os-release | awk -F= '{print $2}' | tr -d '"')
 else
     linux_distro="Unknown"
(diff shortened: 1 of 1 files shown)


src/aznfswatchdog
```diff
@@ -137,7 +137,7 @@ conntrack -L > /dev/null
 
 # conntrack timewait timeout higher than the TCP timewait timeout value isn't very valuable.
 conntrack_timeo_timew=$(cat /proc/sys/net/netfilter/nf_conntrack_tcp_timeout_time_wait 2>/dev/null)
-if [ $? -eq 0 -a $conntrack_timeo_timew -gt $AZNFS_TIMEWAIT_TIMEOUT ]; then
```
linuxsmiths, line 140: this is actually a correct statement. There are two reasons it might be failing:
1. Older versions of bash may by not handling it fine. I don't know about this, but this is a possibility.
2. The second possibility is that one of $conntrack_timeo_timew and $AZNFS_TIMEWAIT_TIMEOUT are null. Since AZNFS_TIMEWAIT_TIMEOUT is defined explicitly, this means conntrack_timeo_timew could be the one undefined, which means the above statement is yielding an empty value. This means netfilter module is not loaded and/or not present. In this case even the current code that you updated will also fail.

let's discuss this

src/aznfswatchdog
```diff
@@ -129,6 +129,17 @@ reconcile_conntrack()
...
+if [ -r /etc/os-release ]; then
+    linux_distro=$(grep "^PRETTY_NAME=" /etc/os-release | awk -F= '{print $2}' | tr -d '"')
+else
+    linux_distro="Unknown"
+fi
+
+bash_version=$(bash --version | head -n 1)
+vecho "Linux distribution: $linux_distro"
+vecho "Bash version: $bash_version"
+
```
cagrawal-msft, line 142: Do all this after L143

src/aznfswatchdog
```diff
@@ -129,6 +129,17 @@ reconcile_conntrack()
     done
 }
 
+# Detect and log distro and bash version
+if [ -r /etc/os-release ]; then
```
cagrawal-msft, line 133: This file is usually not present in centos. Please follow the way it is done in aznfs_install.sh as previously suggested.

src/aznfswatchdog
```diff
@@ -131,13 +131,26 @@ reconcile_conntrack()
 
 vecho "Starting aznfswatchdog..."
 
+# Detect and log distro and bash version
+if [ -f /etc/centos-release ]; then
+    linux_distro=$(cat /etc/centos-release)
```
cagrawal-msft, line 136: If /etc/centos-release is present then it is centos only. Theerefore no need to do cat and just assign the value.

src/aznfswatchdog
```diff
@@ -131,13 +131,26 @@ reconcile_conntrack()
...
+elif [ -f /etc/os-release ]; then
+    linux_distro=$(grep "^PRETTY_NAME=" /etc/os-release | awk -F= '{print $2}' | tr -d '"')
```
cagrawal-msft, line 138: how does the values look for "Pretty_Name" in this file? I used "ID" to get name of the distro before. Please verify this for all distro.

palashvij-msft: Pretty_Name is I think more complete data point, it's combined of Name of distro + version.

for example:

NAME="Rocky Linux"
VERSION="8.7 (Green Obsidian)"
ID="rocky"
ID_LIKE="rhel centos fedora"
VERSION_ID="8.7"
PLATFORM_ID="platform:el8"
PRETTY_NAME="Rocky Linux 8.7 (Green Obsidian)"

src/aznfswatchdog
palashvij-msft: To get complete version number. 
example: 
]# cat /etc/centos-release
CentOS Linux release 8.5.2111

linuxsmiths, line 136: linux_distro=$(cat /etc/centos-release 2>&1)

to direct error to the variable in case there's an error in reading, helps to know that

src/aznfswatchdog
```diff
@@ -131,13 +131,26 @@ reconcile_conntrack()
...
+vecho "Bash version: $bash_version"
+
 # Dump NAT table once on startup in case we have reported conflicts.
 vecho "NAT table:\n$(iptables-save -t nat)"
 conntrack -L > /dev/null
 
 # conntrack timewait timeout higher than the TCP timewait timeout value isn't very valuable.
 conntrack_timeo_timew=$(cat /proc/sys/net/netfilter/nf_conntrack_tcp_timeout_time_wait 2>/dev/null)
-if [ $? -eq 0 -a $conntrack_timeo_timew -gt $AZNFS_TIMEWAIT_TIMEOUT ]; then
+if [ $? -eq 0 ] && [ $conntrack_timeo_timew -gt $AZNFS_TIMEWAIT_TIMEOUT ]; then
```
linuxsmiths, line 153: let's do it like this
if [ $? -eq 0 ] && [ -n "$conntrack_timeo_timew" -a $conntrack_timeo_timew -gt $AZNFS_TIMEWAIT_TIMEOUT ]; then

basically there were some older versions of kernel where this particular setting may not be present hence I think that error might be seen.
Best to check if $conntrack_timeo_timew is a valid value and only then we should check it and try to set it

linuxsmiths, line 153: nit, to be very safe let's do this
if [ $? -eq 0 ] && [ -n "$conntrack_timeo_timew" -a $conntrack_timeo_timew -gt $AZNFS_TIMEWAIT_TIMEOUT ]; then

linuxsmiths (approved):

PR #80, opened 2023-09-11
Fixed syntax error in aznfswatchdog

tirzaelise:
A syntax error in `aznfswatchdog` prevents the script from running, resulting in the inability to mount blob storages.


berend-jutte:
@palashvij-msft Can you check this PR? The syntax error is a big issue for us. We need to add this change to our infra asap. Thanks!

palashvij-msft:
> @palashvij-msft Can you check this PR? The syntax error is a big issue for us. We need to add this change to our infra asap. Thanks!

Hi,
Thanks for pointing out the issue.
We have fixed this in our new release 1.0.10. Please try with that, and Let us know if you face any issues now.

Thanks
Palash

PR #84, opened 2023-09-22
Azure/personal/auto update package

palashvij-msft:


src/update.sh
```diff
@@ -0,0 +1,273 @@
+#!/bin/bash
+
```
linuxsmiths, line 2: Remove this file from the PR.

scripts/aznfs_install.sh
```diff
@@ -4,29 +4,52 @@
...
 apt_update_done=false
 yum="yum"
 apt=0
 zypper=0
 distro_id=
 install_cmd=
 
+# Define the path to the configuration file
+CONFIG_FILE="${OPTDIRDATA}/config"
+
```
linuxsmiths, line 29: nit, move this to L11 right after LOGFILE

scripts/aznfs_install.sh
```diff
@@ -4,29 +4,52 @@
...
+LOGFILE="${OPTDIRDATA}/${APPNAME}.log"
+
 RELEASE_NUMBER=x.y.z
+REPO_OWNER="Azure"
+REPO_NAME="AZNFS-mount"
 AZNFS_RELEASE="aznfs-${RELEASE_NUMBER}-1"
 AZNFS_RELEASE_SUSE="aznfs_sles-${RELEASE_NUMBER}-1"
+SERVICE_NAME="manual-update"
+AUTO_UPDATE_AZNFS=false
+user_wants_update=false
```
linuxsmiths, line 19: nit, let's not define a new variable, we can use AUTO_UPDATE_AZNFS directly

scripts/aznfs_install.sh
```diff
@@ -4,29 +4,52 @@
...
 YELLOW="\e[2;33m"
 NORMAL="\e[0m"
 
+HOSTNAME=$(hostname)
+
 #
 # Core logging function.
 #
-_log()
+_log()
```
linuxsmiths, line 40: nit, remove space at end of line..
do it for all places

scripts/aznfs_install.sh
```diff
@@ -4,29 +4,52 @@
...
+OPTDIR="/opt/microsoft/${APPNAME}"
+OPTDIRDATA="${OPTDIR}/data"
+LOGFILE="${OPTDIRDATA}/${APPNAME}.log"
+
 RELEASE_NUMBER=x.y.z
+REPO_OWNER="Azure"
+REPO_NAME="AZNFS-mount"
 AZNFS_RELEASE="aznfs-${RELEASE_NUMBER}-1"
 AZNFS_RELEASE_SUSE="aznfs_sles-${RELEASE_NUMBER}-1"
+SERVICE_NAME="manual-update"
```
linuxsmiths, line 17: nit, instead of SERVICE_NAME, better name would be RUN_MODE
so users calls this script with a given run mode and it acts accordingly.

scripts/aznfs_install.sh
```diff
@@ -4,29 +4,52 @@
...
 {
     color=$1
     msg=$2
 
+    if [ "$SERVICE_NAME" == "auto-update" ]; then
+        log_message="$(date -u +"%a %b %d %G %T.%3N") $HOSTNAME $$: ${color}${msg}${NORMAL}"
+        (
+            flock -e 999
+            echo -e "$log_message" >> "$LOGFILE"
+        ) 999<"$LOGFILE"
```
linuxsmiths, line 50: is this called from multiple threads?
do we need the flock?

Note for me: Check it

scripts/aznfs_install.sh
```diff
@@ -75,6 +118,31 @@ use_dnf_or_yum()
...
+            return
+        elif [ "${v1_parts[i]}" -gt "${v2_parts[i]}" ]; then
+            echo "-1" # current version > latest_version
+            return
+        fi
+    done
+
+    # If all components are equal, the versions are equal
+    echo "0"
+}
```
linuxsmiths, line 144: we can simply do

is_new_version_available()
{
    local current_version_integer=$(echo $RELEASE_NUMBER | tr -d ".")
    local latest_version_integer=$(echo $1 | tr -d ".")

    [ $current_version_integer -gt $latest_version_integer ]
}

also note that we should only do what's really needed. We don't need to do a full compare (less/greater/equal), all we need to find if the given version is newer, so just do that.

scripts/aznfs_install.sh
```diff
@@ -92,6 +160,19 @@ canonicalize_distro_id()
...
+    local value=$(echo "$1" | tr '[:upper:]' '[:lower:]')
+    
+    if [ "$value" == "true" ] || [ "$value" == "false" ]; then
+        echo "$value"
+    else
+        # If set anything other than boolean values, set false.
+        echo "false"
+    fi
+}
+
```
linuxsmiths, line 175: we don't need this, we can do a case-insensitive compare 

# Convert to lowercase for easy comparison later.
AUTO_UPDATE_AZNFS=${AUTO_UPDATE_AZNFS,,}

scripts/aznfs_install.sh
```diff
@@ -138,7 +219,35 @@ verify_super_user()
     fi
 }
 
-if [ "$RELEASE_NUMBER" == "x.y.z" ]; then
+check_config_file()
```
linuxsmiths, line 222: parse_user_config

scripts/aznfs_install.sh
```diff
@@ -138,7 +219,35 @@ verify_super_user()
...
+
+    pecho "AUTO_UPDATE_AZNFS is set to: $AUTO_UPDATE_AZNFS"
+    
+    # Check if the user has set the environment variable to true
+    if [ "$AUTO_UPDATE_AZNFS" = "true" ]; then
+        user_wants_update=true
+    fi
+}
+
+
```
linuxsmiths, line 243: nit, only leave one empty line

scripts/aznfs_install.sh
```diff
@@ -183,73 +292,184 @@ esac
 
 ensure_pkg "wget"
 
+if [ "$SERVICE_NAME" == "auto-update" ]; then
+    check_config_file
+
+    # Define the GitHub API URL to get the latest release
+    API_URL="https://api.github.com/repos/$REPO_OWNER/$REPO_NAME/releases/latest"
+    # RELEASE_NUMBER=$(curl -s "$API_URL" | grep "tag_name" | cut -d '"' -f 4)
+    RELEASE_NUMBER="0.1.192"
```
linuxsmiths, line 301: remove the hardcode

scripts/aznfs_install.sh
```diff
@@ -183,73 +292,184 @@ esac
...
+        wget "https://github.com/Azure/AZNFS-mount/releases/download/${RELEASE_NUMBER}/${AZNFS_RELEASE}_amd64.deb" -P /tmp
+        apt install -y "/tmp/${AZNFS_RELEASE}_amd64.deb"
+        install_error=$?
+        rm -f "/tmp/${AZNFS_RELEASE}_amd64.deb"
+
+        if [ "$SERVICE_NAME" == "auto-update" ] && [ "$install_error" -eq 0 ]; then
+            pecho "AZNFS updates installed. Restarting aznfswatchdog to apply changes!"
+            systemctl daemon-reload
+            systemctl restart aznfswatchdog
         fi
```
linuxsmiths, line 359: can't this be refactored into a function instead of repeating the same for both distros

src/aznfswatchdog
```diff
@@ -176,6 +189,19 @@ while :; do
...
+    # To check if any AZNFS updates available.
+    # check_for_latest_update=false
+    # if [ $epoch_now -ge $next_update_detection_epoch ]; then
+    #     check_for_latest_update=true
+    #     next_update_detection_epoch=$(expr $(date +%s) + $UPDATE_DETECTION_FREQUENCY)
+    # fi
+
+    # if $check_for_latest_update; then
+    #     ./opt/microsoft/aznfs/aznfs_install.sh
+    # fi
```
linuxsmiths, line 202: can you finalize this

package.sh
```diff
@@ -31,6 +31,7 @@ generate_rpm_package()
 	mkdir -p ${STG_DIR}/${rpm_dir}/tmp${rpm_buildroot_dir}/${rpm_pkg_dir}${opt_dir}
 	cp -avf ${SOURCE_DIR}/lib/common.sh ${STG_DIR}/${rpm_dir}/tmp${rpm_buildroot_dir}/${rpm_pkg_dir}${opt_dir}/
 	cp -avf ${SOURCE_DIR}/src/mountscript.sh ${STG_DIR}/${rpm_dir}/tmp${rpm_buildroot_dir}/${rpm_pkg_dir}${opt_dir}/
+    cp -avf ${SOURCE_DIR}/scripts/aznfs_install.sh ${STG_DIR}/${rpm_dir}/tmp${rpm_buildroot_dir}/${rpm_pkg_dir}${opt_dir}/
```
linuxsmiths, line 34: indent

packaging/aznfs/DEBIAN/postinst
```diff
@@ -38,7 +39,25 @@ if [ -f /opt/microsoft/aznfs/randbytes ]; then
...
+
+# Check if the config file exists; if not, create it.
+if [ ! -f /opt/microsoft/aznfs/data/config ]; then
+        # Create the config file and set default AUTO_UPDATE_AZNFS=false inside it.
+        echo "AUTO_UPDATE_AZNFS=false" > /opt/microsoft/aznfs/data/config
+
+        # Set the permissions for the config file.
+        chmod 0644 /opt/microsoft/aznfs/data/config
+else
+        chattr -f -i /opt/microsoft/aznfs/data/config
```
linuxsmiths, line 51: remove this.. in general don't try to do anything with this file, if present we should just read it and get the config values, else use the default values

packaging/aznfs/DEBIAN/postinst
```diff
@@ -38,7 +39,25 @@ if [ -f /opt/microsoft/aznfs/randbytes ]; then
...
-# Start aznfswatchdog service.
-systemctl daemon-reload
-systemctl enable aznfswatchdog
-systemctl start aznfswatchdog
\ No newline at end of file
+
+# Check if the config file exists; if not, create it.
+if [ ! -f /opt/microsoft/aznfs/data/config ]; then
+        # Create the config file and set default AUTO_UPDATE_AZNFS=false inside it.
+        echo "AUTO_UPDATE_AZNFS=false" > /opt/microsoft/aznfs/data/config
```
linuxsmiths, line 46: let's make the default true

packaging/aznfs/DEBIAN/postinst
```diff
@@ -38,7 +39,25 @@ if [ -f /opt/microsoft/aznfs/randbytes ]; then
...
+
+        # Set the permissions for the config file.
+        chmod 0644 /opt/microsoft/aznfs/data/config
+else
+        chattr -f -i /opt/microsoft/aznfs/data/config
+fi
+
```
linuxsmiths, line 53: aznfs_install.sh can put the aznfswatchdog pid in this file and here we can check if it matches, if not it's a stale file and delete and proceed.

packaging/aznfs/DEBIAN/postrm
```diff
@@ -6,7 +6,9 @@
...
 if [ "$1" == 'remove' -o "$1" == 'purge' ]; then
-	chattr -f -i /opt/microsoft/aznfs/data/mountmap
-	chattr -f -i /opt/microsoft/aznfs/data/randbytes
-	rm -rf /opt/microsoft/aznfs
+    chattr -f -i /opt/microsoft/aznfs/data/mountmap
+    chattr -f -i /opt/microsoft/aznfs/data/randbytes
+    rm -rf /opt/microsoft/aznfs
+elif [ "$1" == 'upgrade' ]; then
+    chattr -f +i /opt/microsoft/aznfs/data/config
 fi
```
linuxsmiths, line 14: remove this

scripts/aznfs_install.sh
```diff
@@ -75,6 +115,83 @@ use_dnf_or_yum()
...
+check_and_perform_update()
+{
+    local current_version="$1"
+
+    # Check if the service name is "auto-update"
+    if [ "$RUN_MODE" == "auto-update" ]; then
+        # Compare the current version with the latest release
+        if is_new_version_available "$current_version" "$RELEASE_NUMBER"; then
+            # Check if an update is available
+            if [ "$AUTO_UPDATE_AZNFS" = "true" ]; then
```
linuxsmiths, line 152: let's use == consistently

packaging/aznfs/DEBIAN/postinst
```diff
@@ -38,7 +39,23 @@ if [ -f /opt/microsoft/aznfs/randbytes ]; then
...
+# Check if the config file exists; if not, create it.
+if [ ! -f /opt/microsoft/aznfs/data/config ]; then
+        # Create the config file and set default AUTO_UPDATE_AZNFS=true inside it.
+        echo "AUTO_UPDATE_AZNFS=true" > /opt/microsoft/aznfs/data/config
+
+        # Set the permissions for the config file.
+        chmod 0644 /opt/microsoft/aznfs/data/config
+fi
+
+# Check if the flag file does not exist
```
linuxsmiths, line 52: more useful comment would be

If it's an auto upgrade triggered by aznfswatchdog, don't restart watchdog.

linuxsmiths: also, let's make the file hidden with name starting with a dot
/tmp/.update_in_progress_from_watchdog.flag

packaging/aznfs/DEBIAN/postrm
```diff
@@ -6,7 +6,7 @@
 # --------------------------------------------------------------------------------------------
 
 if [ "$1" == 'remove' -o "$1" == 'purge' ]; then
-	chattr -f -i /opt/microsoft/aznfs/data/mountmap
-	chattr -f -i /opt/microsoft/aznfs/data/randbytes
-	rm -rf /opt/microsoft/aznfs
+    chattr -f -i /opt/microsoft/aznfs/data/mountmap
+    chattr -f -i /opt/microsoft/aznfs/data/randbytes
+    rm -rf /opt/microsoft/aznfs
```
linuxsmiths, line 11: undo this space change

packaging/aznfs/DEBIAN/prerm
```diff
@@ -26,7 +26,26 @@ if [ "$1" == 'remove' -o "$1" == 'purge' ]; then
...
-echo "Stopped aznfswatchdog service."
\ No newline at end of file
+flag_file="/tmp/update_in_progress_from_watchdog.flag"
+
+if [ -f "$flag_file" ]; then
+	# Get the PID of aznfswatchdog
+	aznfswatchdog_pid=$(pgrep aznfswatchdog)
+	
+	# Read the PID from the flag file
+	aznfswatchdog_pid_inside_flag=$(cat "$flag_file")
```
linuxsmiths, line 36: nit, don't need quotes around $flag_file

scripts/aznfs_install.sh
```diff
@@ -75,6 +115,93 @@ use_dnf_or_yum()
...
+            package_name=${AZNFS_RELEASE}_amd64.deb
+        elif [ "$install_cmd" == "zypper" ]; then
+            AZNFS_RELEASE_SUSE="aznfs_sles-${RELEASE_NUMBER}-1"
+            package_name=${AZNFS_RELEASE_SUSE}.x86_64.rpm
+        else
+            AZNFS_RELEASE="aznfs-${RELEASE_NUMBER}-1"
+            package_name=${AZNFS_RELEASE}.x86_64.rpm
+        fi
+        
+        wget "https://github.com/Azure/AZNFS-mount/releases/download/${RELEASE_NUMBER}/${package_name}" -P /tmp
```
linuxsmiths, line 142: wget can likely fail, so we need to check the return status and also check for presence of a valid "/tmp/${package_name}" before going ahead with installation.
If wget fails, a useful error log would help.

scripts/aznfs_install.sh
```diff
@@ -75,6 +115,93 @@ use_dnf_or_yum()
...
+            $install_cmd install -y "/tmp/${package_name}"
+        fi
+        install_error=$?
+        rm -f "/tmp/${package_name}"
+
+        if [ "$RUN_MODE" == "auto-update" ] && [ "$install_error" -eq 0 ]; then
+            pecho "AZNFS updates installed. Restarting aznfswatchdog to apply changes!"
+            systemctl daemon-reload
+            systemctl restart aznfswatchdog
+            exit 1 # Nothing in the script will run after this point.
```
linuxsmiths, line 155: don't exit with 1 since it's a successful exit
exit 0

scripts/aznfs_install.sh
```diff
@@ -75,6 +115,93 @@ use_dnf_or_yum()
...
+            # Check if an update is available
+            if [ "$AUTO_UPDATE_AZNFS" == "true" ]; then   
+                # Get the PID of aznfswatchdog
+                aznfswatchdog_pid=$(pgrep aznfswatchdog)
+                pecho "aznfswatchdog_pid: $aznfswatchdog_pid" # remove later
+                if [ -n "$aznfswatchdog_pid" ]; then
+                    # Create a flag file with the PID to indicate that an update is in progress
+                    echo "$aznfswatchdog_pid" > /tmp/update_in_progress_from_watchdog.flag
+                else
+                    eecho "aznfswatchdog process not found."
```
linuxsmiths, line 177: Instead can you say
AZNFS auto-update can only be invoked by aznfswatchdog!

scripts/aznfs_install.sh
```diff
@@ -183,73 +336,48 @@ esac
 
 ensure_pkg "wget"
 
+if [ "$RUN_MODE" == "auto-update" ]; then
+    parse_config_file
+
+    # Define the GitHub API URL to get the latest release
+    API_URL="https://api.github.com/repos/$REPO_OWNER/$REPO_NAME/releases/latest"
+    # RELEASE_NUMBER=$(curl -s "$API_URL" | grep "tag_name" | cut -d '"' -f 4)
+    RELEASE_NUMBER="0.1.199"
```
linuxsmiths, line 345: this needs to be uncommented.
Once you uncomment it, then you should also check that the above pipeline yielded a valid RELEASE_NUMBER.
It might be better to make the curl separate so that we can check the return status of that as that's a n/w operation and can likely fail. If it fails, we should not proceed

scripts/aznfs_install.sh
```diff
@@ -183,73 +336,48 @@ esac
 
 ensure_pkg "wget"
 
+if [ "$RUN_MODE" == "auto-update" ]; then
+    parse_config_file
+
+    # Define the GitHub API URL to get the latest release
+    API_URL="https://api.github.com/repos/$REPO_OWNER/$REPO_NAME/releases/latest"
```
cagrawal-msft, line 343: Also, I remember there is limitation on how many times this github api can be called per day/hour? If that is the case, then we have to think about that as well because it is running in loop and we don't know how many times it will get called.

scripts/aznfs_install.sh
```diff
@@ -75,6 +115,93 @@ use_dnf_or_yum()
...
+        if [ "$RUN_MODE" == "auto-update" ] && [ "$install_error" -eq 0 ]; then
+            pecho "AZNFS updates installed. Restarting aznfswatchdog to apply changes!"
+            systemctl daemon-reload
+            systemctl restart aznfswatchdog
+            exit 1 # Nothing in the script will run after this point.
+        fi
+    fi
+}
+
+check_and_perform_update()
```
cagrawal-msft, line 160: In this function, we are not issuing the update command and the name is conflicting with the other perform_aznfs_update function. We can change the name to just "check_aznfs_update"

scripts/aznfs_install.sh
palashvij-msft: I have read about that, the rate-limit for github API for unauntheticated user is 60 times/hour. I have also verified that. Furthermore, it's okay for us to use this API as we'll be setting the frequency to check for updates once a day or once a week. We'll be running it in a loop, but the condition to run the update script will be true only once.

cagrawal-msft: If the limit is for per user, then it is fine. I thought total number of times this API can be called for a repo is restricted.

package.sh
```diff
@@ -87,8 +88,14 @@ generate_tarball_package() {
 
     # Copy other static package file(s).
     mkdir -p ${STG_DIR}/tarball/${tar_pkg_dir}/usr/sbin
+	
+    # Set AKS_USER variable to true inside aznfswatchdog to indicate use by Azure Kubernetes Service (AKS)
+    sed -i 's/AKS_USER=false/AKS_USER=true/' ${SOURCE_DIR}/src/aznfswatchdog
```
cagrawal-msft, line 93: Instead of changing the var in the original file, update the value once it is copied to specific location from where it will be included in tarball. 

In short, do this sed after L94 for "${STG_DIR}/tarball/${tar_pkg_dir}/usr/sbin/aznfswatchdog". Then you won't nedd L97 as well.

packaging/aznfs/DEBIAN/postinst
```diff
@@ -38,7 +39,23 @@ if [ -f /opt/microsoft/aznfs/randbytes ]; then
...
+if [ ! -f /opt/microsoft/aznfs/data/config ]; then
+        # Create the config file and set default AUTO_UPDATE_AZNFS=true inside it.
+        echo "AUTO_UPDATE_AZNFS=true" > /opt/microsoft/aznfs/data/config
+
+        # Set the permissions for the config file.
+        chmod 0644 /opt/microsoft/aznfs/data/config
+fi
+
+# If it's an auto upgrade triggered by aznfswatchdog, don't restart watchdog.
+if [ ! -f /tmp/.update_in_progress_from_watchdog.flag ]; then
```
cagrawal-msft, line 53: I see you have used falg_file var for this file name in other files. Since it is used here as well multiple times, you can use the same var name here as well to be consistent.

packaging/aznfs/DEBIAN/prerm
```diff
@@ -26,7 +26,25 @@ if [ "$1" == 'remove' -o "$1" == 'purge' ]; then
...
+	aznfswatchdog_pid_inside_flag=$(cat $flag_file)
+	
+	if [ "$aznfswatchdog_pid" != "$aznfswatchdog_pid_inside_flag" ]; then
+		# The flag file is stale, remove it
+		rm -f "$flag_file"
+		echo "Removed stale flag file."
+	fi
+fi
+
+if [ ! -f "$flag_file" ]; then
```
cagrawal-msft, line 45: This is else condition of above if condition, right? Then write it in proper if else.

src/aznfswatchdog
```diff
@@ -176,6 +182,22 @@ while :; do
...
 
+    # To check if any AZNFS updates available.
+    check_for_latest_update=false
+    if [ "$AKS_USER" != "true" ] && [ $epoch_now -ge $next_update_detection_epoch ]; then
+        check_for_latest_update=true
+        next_update_detection_epoch=$(expr $(date +%s) + $UPDATE_DETECTION_FREQUENCY)
+    fi
+
+    if $check_for_latest_update; then
+        if [ -e "/opt/microsoft/aznfs/aznfs_install.sh" ]; then
```
cagrawal-msft, line 193: Add a var for this file as well in common.sh and use that shorten var everywhere.

scripts/aznfs_install.sh
```diff
@@ -183,73 +349,73 @@ esac
...
+
+    # Define the GitHub API URL to get the latest release
+    API_URL="https://api.github.com/repos/$REPO_OWNER/$REPO_NAME/releases/latest"
+    RELEASE_INFO=$(curl -s "$API_URL")
+    if [ $? -ne 0 ]; then
+        eecho "Failed to retrieve latest release information. Exiting."
+        exit 1
+    fi
+
+    # Parse the release number from the JSON response
```
cagrawal-msft, line 363: Full stop at the end of each comment.

scripts/aznfs_install.sh
```diff
@@ -183,73 +349,73 @@ esac
...
+    RELEASE_NUMBER=$(echo "$RELEASE_INFO" | grep "tag_name" | cut -d '"' -f 4)
+    if [ -z "$RELEASE_NUMBER" ]; then
+        eecho "Failed to retrieve latest release number from Latest release information. Exiting."
```
cagrawal-msft, line 366: nit: small L for second latest.

packaging/aznfs/DEBIAN/prerm
```diff
@@ -26,7 +26,25 @@ if [ "$1" == 'remove' -o "$1" == 'purge' ]; then
...
+	# Get the PID of aznfswatchdog
+	aznfswatchdog_pid=$(pgrep aznfswatchdog)
+	
+	# Read the PID from the flag file
+	aznfswatchdog_pid_inside_flag=$(cat $flag_file)
+	
+	if [ "$aznfswatchdog_pid" != "$aznfswatchdog_pid_inside_flag" ]; then
+		# The flag file is stale, remove it
+		rm -f "$flag_file"
+		echo "Removed stale flag file."
```
cagrawal-msft, line 41: Instead of saying just flag file, you should add path of the file.

packaging/aznfs/DEBIAN/prerm
```diff
@@ -26,7 +26,25 @@ if [ "$1" == 'remove' -o "$1" == 'purge' ]; then
...
-# Stop aznfswatchdog service and clean it.
-systemctl stop aznfswatchdog
-systemctl disable aznfswatchdog
-echo "Stopped aznfswatchdog service."
\ No newline at end of file
+flag_file="/tmp/.update_in_progress_from_watchdog.flag"
+
+if [ -f "$flag_file" ]; then
+	# Get the PID of aznfswatchdog
+	aznfswatchdog_pid=$(pgrep aznfswatchdog)
```
cagrawal-msft, line 33: I guess better var name should be pid_aznfswatchdog.

packaging/aznfs/DEBIAN/prerm
```diff
@@ -26,7 +26,25 @@ if [ "$1" == 'remove' -o "$1" == 'purge' ]; then
...
+	
+	# Read the PID from the flag file
+	aznfswatchdog_pid_inside_flag=$(cat $flag_file)
```
cagrawal-msft, line 36: better var name will be pid_aznfswatchdog_in_flagfile

packaging/aznfs/DEBIAN/prerm
palashvij-msft: we have already kept file hidden, I think we should not print that. we can discuss on this. file is already made hidden

package.sh
```diff
@@ -88,6 +89,9 @@ generate_tarball_package() {
     # Copy other static package file(s).
     mkdir -p ${STG_DIR}/tarball/${tar_pkg_dir}/usr/sbin
     cp -avf ${SOURCE_DIR}/src/aznfswatchdog ${STG_DIR}/tarball/${tar_pkg_dir}/usr/sbin
+	
+	# Set AKS_USER variable to true inside aznfswatchdog to indicate use by Azure Kubernetes Service (AKS)
+	sed -i 's/AKS_USER="false"/AKS_USER="true"/' ${STG_DIR}/tarball/${tar_pkg_dir}/usr/sbin/aznfswatchdog
```
cagrawal-msft, line 94: Indentation

linuxsmiths (approved):

packaging/aznfs/DEBIAN/postinst
cagrawal-msft: @palashvij-msft this one is still remaining.

scripts/aznfs_install.sh
```diff
@@ -75,6 +115,111 @@ use_dnf_or_yum()
...
+
+    if [ $install_error -ne 0 ]; then
+        eecho "[FATAL] Error installing AZNFS $RELEASE_NUMBER (Error: $install_error). See '$install_cmd' command logs for more information"
+        eecho "$install_output"
+        exit 1
+    fi
+
+    if [ "$RUN_MODE" == "auto-update" ]; then
+        secho "Successfully updated AZNFS version $current_version to $RELEASE_NUMBER"
+        pecho "Restarting aznfswatchdog to apply changes!"
```
cagrawal-msft, line 170: nit: Don't end success logs with !, specific to this one you can add "..."

scripts/aznfs_install.sh
```diff
@@ -75,6 +115,111 @@ use_dnf_or_yum()
...
+                eecho "AZNFS auto-update can only be invoked by aznfswatchdog!"
+                exit 1
+            fi
+        else
+            pecho "Not auto-updating to $RELEASE_NUMBER. AZNFS version $current_version is up-to-date or newer!"
+            exit 0
+        fi
+        
+    elif [ "$RUN_MODE" == "manual-update" ]; then
+        # Check if the current version matches the desired release number
```
cagrawal-msft, line 207: nit: Fulll stop. 
at L213 as well.

package.sh
```diff
@@ -89,6 +90,9 @@ generate_tarball_package() {
     mkdir -p ${STG_DIR}/tarball/${tar_pkg_dir}/usr/sbin
     cp -avf ${SOURCE_DIR}/src/aznfswatchdog ${STG_DIR}/tarball/${tar_pkg_dir}/usr/sbin
 
+    # Set AKS_USER variable to true inside aznfswatchdog to indicate use by Azure Kubernetes Service (AKS)
```
cagrawal-msft, line 93: nit: . at end.

packaging/aznfs/DEBIAN/postinst
```diff
@@ -38,7 +39,23 @@ if [ -f /opt/microsoft/aznfs/randbytes ]; then
...
+fi
+flag_file="/tmp/.update_in_progress_from_watchdog.flag"
+
+# If it's an auto update triggered by aznfswatchdog, don't restart watchdog.
+if [ ! -f "$flag_file" ]; then
+        systemctl daemon-reload
+        systemctl enable aznfswatchdog
+        systemctl start aznfswatchdog
+else
+        # Clean up the update in progress flag file
```
cagrawal-msft, line 58: nit: . at end.

cagrawal-msft: In other files as well.

scripts/aznfs_install.sh
```diff
@@ -75,6 +115,111 @@ use_dnf_or_yum()
...
+    rm -f "/tmp/${package_name}"
+
+    if [ $install_error -ne 0 ]; then
+        eecho "[FATAL] Error installing AZNFS $RELEASE_NUMBER (Error: $install_error). See '$install_cmd' command logs for more information"
+        eecho "$install_output"
+        exit 1
+    fi
+
+    if [ "$RUN_MODE" == "auto-update" ]; then
+        secho "Successfully updated AZNFS version $current_version to $RELEASE_NUMBER..."
```
cagrawal-msft, line 169: I think you misunderstood. "..." should be added in case if some process is running. f.e. in line just below this. Please update.

scripts/aznfs_install.sh
```diff
@@ -75,6 +115,111 @@ use_dnf_or_yum()
...
+    if [ "$install_cmd" == "zypper" ]; then
+        install_output=$($install_cmd install --allow-unsigned-rpm -y "/tmp/${package_name}" 2>&1)
+    else
+        install_output=$($install_cmd install -y "/tmp/${package_name}" 2>&1)
+    fi
+    install_error=$?
+    rm -f "/tmp/${package_name}"
+
+    if [ $install_error -ne 0 ]; then
+        eecho "[FATAL] Error installing AZNFS $RELEASE_NUMBER (Error: $install_error). See '$install_cmd' command logs for more information"
```
cagrawal-msft, line 163: Since we have used "AZNFS version $RELEASE_NUMBER" everywhere else, would be good to keep it consistent.

scripts/aznfs_install.sh
```diff
@@ -138,7 +283,40 @@ verify_super_user()
...
+{
+    if [ -f "$CONFIG_FILE" ]; then
+        # Read the value of AUTO_UPDATE_AZNFS from the configuration file.
+        AUTO_UPDATE_AZNFS=$(grep "^AUTO_UPDATE_AZNFS=" "$CONFIG_FILE" | cut -d '=' -f2)
+        
+        # Convert to lowercase for easy comparison later.
+        AUTO_UPDATE_AZNFS=${AUTO_UPDATE_AZNFS,,}
+    else
+        # If the configuration file doesn't exist, set a default value.
+        eecho "$CONFIG_FILE not found. Setting the default values!"
```
cagrawal-msft, line 296: Setting the default value!
Since we are setting just one value.

scripts/aznfs_install.sh
```diff
@@ -138,7 +283,40 @@ verify_super_user()
     fi
 }
 
-if [ "$RELEASE_NUMBER" == "x.y.z" ]; then
+parse_user_config()
+{
+    if [ -f "$CONFIG_FILE" ]; then
+        # Read the value of AUTO_UPDATE_AZNFS from the configuration file.
+        AUTO_UPDATE_AZNFS=$(grep "^AUTO_UPDATE_AZNFS=" "$CONFIG_FILE" | cut -d '=' -f2)
```
cagrawal-msft, line 290: @linuxsmiths we are not handling the case if config file is present but is empty or AUTO_UPDATE_AZNFS is not present? 
@palashvij-msft we should handle this case as well.

linuxsmiths (approved):

PR #89, opened 2023-11-16
handling the hung issue with RA

palashvij-msft:
adding logic to delete unreplied conntrack entries and resilient to frequent IP changes as seen in case of RA

### DIFF
--- a/src/aznfswatchdog
+++ b/src/aznfswatchdog
@@ -17,8 +17,8 @@ MONITOR_INTERVAL_SECS=5
 # How often do we check for change in FQDN->IP?
 IP_CHANGE_DETECTION_FREQUENCY=60
 
-# How often do we check for updates of AZNFS?
-UPDATE_DETECTION_FREQUENCY=86400
+# Determines the frequency at which we check for updates of AZNFS and run the daily function.
+DAILY_CHECK_AFTER_SECONDS=86400
 
 #
 # Remove unmounted entries only if MOUNTMAP has not been changed till MONITOR_INTERVAL_SECS seconds.
@@ -80,7 +80,7 @@ declare -A regional_accounts
 should_skip_regional_account()
 {
     local l_host=$1
-    
+
     if [ ${regional_accounts["$l_host"]} -eq 1 ]; then
         # Environment variable prohibits skipping, return failure to do so.
         if [ $AZNFS_SKIP_IP_CHANGE_DETECTION_FOR_REGIONAL_ACCOUNTS -eq 0 ]; then
@@ -97,6 +97,10 @@ should_skip_regional_account()
 check_for_regional_account()
 {
     local l_host=$1
+
+    if [ $AZNFS_SKIP_IP_CHANGE_DETECTION_FOR_REGIONAL_ACCOUNTS -eq 0 ]; then
+        return 0
+    fi
 
     let ip_change_count[$l_host]++
     local current_time=$(date +"%s")
@@ -283,7 +287,7 @@ ping_new_endpoint()
 
 run_daily_once()
 {
-    next_daily_run_epoch=$(expr $(date +%s) + $UPDATE_DETECTION_FREQUENCY)
+    next_daily_run_epoch=$(expr $(date +%s) + $DAILY_CHECK_AFTER_SECONDS)
     
     #
     # To increase resilience in case the account is marked as regional due to a bad state,
--- a/src/aznfswatchdog
+++ b/src/aznfswatchdog
@@ -368,7 +368,9 @@ if ! chattr -f +i $MOUNTMAP; then
     wecho "chattr does not work for ${MOUNTMAP}!"
 fi
 
-
+if [ $AZNFS_SKIP_IP_CHANGE_DETECTION_FOR_REGIONAL_ACCOUNTS -eq 0]; then
+    MOUNTMAP_INACTIVITY_SECS=45
+fi
 
 #
 # Watchdog for monitoring unmounts and more importantly change in blob endpoint
--- a/src/aznfswatchdog
+++ b/src/aznfswatchdog
@@ -120,7 +120,7 @@ reconcile_conntrack_unreplied()
     # We are called every 5 secs, so this deletes an entry stuck in
     # UNREPLIED for 25/30 secs.
     #
-    if [ ${cthash_synsent[$key]} -ge 5 ]; then
+    if [ ${cthash_unreplied[$key]} -ge 5 ]; then
         cmd="conntrack -D -p tcp -d $l_ip -r $l_reply_srcip --sport $l_sport --dport $l_dport"
         wecho "Deleting conntrack entry stuck in UNREPLIED state [$cmd]"
 
@@ -167,8 +167,8 @@ reconcile_conntrack()
         done
     fi
 
-    IFS=$'\n' output111=$(conntrack -L -p tcp -d $l_ip --dport 111 | grep "\[UNREPLIED\]" 2>/dev/null)
-    IFS=$'\n' output2048=$(conntrack -L -p tcp -d $l_ip --dport 2048 | grep "\[UNREPLIED\]" 2>/dev/null)
+    IFS=$'\n' output111=$(conntrack -L -p tcp -d $l_ip --dport 111 | grep -v "SYN_SENT" | grep "\[UNREPLIED\]" 2>/dev/null)
+    IFS=$'\n' output2048=$(conntrack -L -p tcp -d $l_ip --dport 2048 | grep -v "SYN_SENT" | grep "\[UNREPLIED\]" 2>/dev/null)
     output="$output111"$'\n'"$output2048"
     if [ -n "$output" ]; then
         for line in $output; do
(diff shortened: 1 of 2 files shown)


src/aznfswatchdog
```diff
@@ -65,33 +65,60 @@ chattr -f +i $RANDBYTES
 # kernels. If we find an entry stuck for more than a certain time in SYN_SENT state
 # we delete the entry so that kernel looks up fresh NAT rules and creates a new entry.
 #
-declare -A cthash_synsent
-declare -A cthash_synsent_now
+declare -A cthash_stuck
+declare -A cthash_stuck_now
```
linuxsmiths, line 69: for clarity, leave the prev ones and add new one for unreplied state tracking

`#`
`#` Hash for storing how many times we have seen a conntrack entry in UNREPLIED state.
`#` Used for finding if some entry is stuck in UNREPLIED state due to no response from
`#` NFS server or some n/w issue. We want to delete the entry as it prevents creation of
`#` a new conntrack entry, if the IP were to change again soon.
`#`
declare -A cthash_unreplied
declare -A cthash_unreplied_now

src/aznfswatchdog
```diff
@@ -65,33 +65,60 @@ chattr -f +i $RANDBYTES
...
 
 reconcile_conntrack_one()
```
linuxsmiths, line 71: rename this to reconcile_conntrack_synsent and add a new one reconcile_conntrack_unreplied

src/aznfswatchdog
```diff
@@ -122,22 +149,57 @@ reconcile_conntrack()
...
                 l_sport=${BASH_REMATCH[1]}
                 l_dport=${BASH_REMATCH[2]}
-                reconcile_conntrack_one $l_ip $l_sport $l_dport $l_nfsip
+                reconcile_conntrack_one $l_ip $l_sport $l_dport $l_nfsip "SYN_SENT"
+            fi
+        done
+    fi
+
+    IFS=$'\n' output111=$(conntrack -L -p tcp -d $l_ip --dport 111 --state ESTABLISHED 2>/dev/null)
+    IFS=$'\n' output2048=$(conntrack -L -p tcp -d $l_ip --dport 2048 --state ESTABLISHED 2>/dev/null)
```
linuxsmiths, line 158: filter UNREPLIED entries only and don't restrict to ESTABLISHED state only. Sometimes if there is some n/w issue it may be in some other state. What we want to clear are entries with dst ip equal to the proxy IP and ports matching 111 and 2048. if such an entry is stuck in UNREPLIED state for more than 30 secs we want to delete it.

    IFS=$'\n' output111=$(conntrack -L -p tcp -d $l_ip --dport 111 | grep "\[UNREPLIED\]" 2>/dev/null)

    IFS=$'\n' output2048=$(conntrack -L -p tcp -d $l_ip --dport 2048 | grep "\[UNREPLIED\]" 2>/dev/null)

src/aznfswatchdog
```diff
@@ -122,22 +149,57 @@ reconcile_conntrack()
...
+    output="$output111"$'\n'"$output2048"
+    if [ -n "$output" ]; then
+        for line in $output; do
+            # XXX Remove this log after running for few days.
+            vecho "$line"
+
+            matchstr=".* ESTABLSIHED src=([0-9]+\.[0-9]+\.[0-9]+\.[0-9]+) dst=$l_ip sport=([0-9]+) dport=([0-9]+) [UNREPLIED].*"
```
linuxsmiths, line 165: again, remove ESTABLSIHED  from here

src/aznfswatchdog
```diff
@@ -122,22 +149,57 @@ reconcile_conntrack()
...
+            if [[ "$line" =~ $matchstr ]]; then
+                l_sysip=${BASH_REMATCH[1]}
```
linuxsmiths, line 167: this is the source IP so this will be the interface IP, hence this is not very interesting. Also u cannot use this at L90 since you use -r there which means source in the reply direction.
What is more interesting is the src after UNREPLIED. This is the source in the reply direction and will be useful for matching.
save that as l_reply_srcip and then you can use it at L90

src/aznfswatchdog
```diff
@@ -65,33 +65,60 @@ chattr -f +i $RANDBYTES
...
-        if [ $? -ne 0 ]; then
-            eecho "Failed to delete conntrack entry [$cmd]!"
-        else
-            unset cthash_synsent[$key]
+        #
+        # We are called every 5 secs, so this deletes an entry stuck in
+        # UNREPLIED for 25/30 secs.
+        #
+        if [ ${cthash_stuck[$key]} -ge 5 ]; then
+            cmd="conntrack -D -p tcp -d $l_ip -r $l_sysip --sport $l_sport --dport $l_dport --state ESTABLISHED"
```
linuxsmiths, line 90: again, remove --state ESTABLISHED, just matching based on the other fields is enough

src/aznfswatchdog
```diff
@@ -122,22 +149,57 @@ reconcile_conntrack()
...
+            unset cthash_stuck[$key]
         fi
     done
 }
 
+#
+# Function for running stat for mountpoint so that everytime DNAT rule is updated, it is 
+# used to make sure to send atleast one packet that matches the DNAT rule.
+# This will eventually make sure the conntrack entry created always for communication.
+#
```
linuxsmiths, line 190: `#`
`#` Function for running stat for mountpoint so that everytime DNAT rule is updated, it is 
`#` used to make sure to send atleast one packet that matches the DNAT rule.
`#` This will make sure that the connection gets a TCP reset and outstanding NFS RPC requests
`#` are retransmitted right away w/o waiting for the 1 min timeout.
`#`

src/aznfswatchdog
```diff
@@ -122,22 +149,57 @@ reconcile_conntrack()
...
+{
+    local target="$1"
+    vecho "[$BASHPID @ $(date)] Starting bg stat"
+    statres=$(stat "$target")
+    vecho "[$BASHPID @ $(date) #1] statres=$statres"
+    sleep 35
+    statres=$(stat "$target")
+    vecho "[$BASHPID @ $(date) #2] statres=$statres"
+    vecho "[$BASHPID @ $(date)] Finishing bg stat"
+}
```
linuxsmiths, line 201: ping_new_endpoint()
{
    local target="$1"

    vecho "[$BASHPID] stat($target) #1 start"
    stat "$target"
    vecho "[$BASHPID] stat($target) #1 done"

    sleep 35
    
    # One more stat after 30 sec sleep to let dir attributes timeout.
    vecho "[$BASHPID] stat($target) #2 start"
    stat "$target"
    vecho "[$BASHPID] stat($target) #2 done"
}

src/aznfswatchdog
```diff
@@ -382,6 +444,10 @@ while :; do
             if ! update_mountmap_entry "$line" "$l_host $l_ip $new_ip"; then
                 eecho "Will reattempt the operation in next iteration."
             fi
+            mountpoint=$(echo "$findmnt" | grep " nfs ${l_ip}:" | awk '{print $4}')
+            run_stat "$mountpoint" &
+            background_pid=$!
+            vecho "Background process started with PID: $background_pid"
```
linuxsmiths, line 450: put this in else
and don't need to print background_pid log

src/aznfswatchdog
```diff
@@ -122,20 +220,80 @@ reconcile_conntrack()
...
+# Function for running stat for mountpoint so that everytime DNAT rule is updated, it is
+# used to make sure to send atleast one packet that matches the DNAT rule.
+# This will make sure that the connection gets a TCP reset and outstanding NFS RPC requests
+# are retransmitted right away w/o waiting for the 1 min timeout.
+#
+ping_new_endpoint()
+{
+    local target="$1"
+
+    stat "$target"
```
linuxsmiths, line 273: why did u remove the useful logging?
these anyways are rare events, so we can log

src/aznfswatchdog
```diff
@@ -122,20 +220,80 @@ reconcile_conntrack()
...
+    #
+    ip_change_count=()
+    last_ip_change_time=()
+    regional_accounts=()
+
+    # Since AKS users don't require auto-update, there's no need to check for updates.
+    if [ "$AKS_USER" == "true" ]; then
+        return 1
+    fi
+    check_for_latest_update=true
```
linuxsmiths, line 296: since this AKS check is specifically for update, better to do

if [ "$AKS_USER" != "true" ]; then
    check_for_latest_update=true
fi

linuxsmiths (approved):

src/aznfswatchdog
```diff
@@ -377,12 +558,18 @@ while :; do
         #
         if [ "$new_ip" != "$l_nfsip" ]; then
             pecho "IP for $l_host changed [$l_nfsip -> $new_ip]."
+            check_for_regional_account "$l_host"
 
             # This will update DNAT rule as well.
             if ! update_mountmap_entry "$line" "$l_host $l_ip $new_ip"; then
                 eecho "Will reattempt the operation in next iteration."
+            else
+                mountpoint=$(echo "$findmnt" | grep " nfs ${l_ip}:" | awk '{print $4}')
```
linuxsmiths, line 567: this breaks if local mount direcctory name has spaces

src/aznfswatchdog
```diff
@@ -59,6 +70,60 @@ if [ ! -s $RANDBYTES ]; then
...
+
+#
+# To check if an account should be marked as a "Regional Account".
+#
+check_for_regional_account()
+{
+    local l_host=$1
+
+    if [[ ${regional_accounts["$l_host"]} -eq 1 ]]; then
+        return 0
```
linuxsmiths, line 104: simply return, since caller doesn't check return type

linuxsmiths (approved):

PR #90, opened 2023-11-27
aznfs Mount Helper for nfsv4

rmirashi-msft:


src/mountscript.sh
```diff
@@ -602,6 +644,324 @@ ensure_aznfswatchdog()
...
+
+    if [ $port -le $NFSV4_PORT_RANGE_END ]; then
+        echo "$port"
+    else
+        echo ""
+    fi
+}
+
+
+install_CA_cert()
```
sharimsft, line 713: we might have to have an option to download or use an existing copy of the certificate

src/mountscript.sh
```diff
@@ -602,6 +644,324 @@ ensure_aznfswatchdog()
...
+
+
+#
+# Add stunnel configuration in stunnel_<storageaccount>.conf file.
+#
+add_stunnel_configuration()
+{
+    chattr -f -i $stunnel_conf_file
+
+    stunnel_CAFile="/etc/ssl/certs/DigiCert_Global_Root_G2.pem"
```
sharimsft, line 732: can this not be a global variable?

src/mountscript.sh
```diff
@@ -217,13 +225,47 @@ get_dir_from_share()
 
     if [ -z "$account" -o -z "$container" -o -n "$extra" ]; then
         eecho "Bad share name: ${hostshare}."
-        eecho "Share to be mounted must be of the form 'account.blob.core.windows.net:/account/container'."
+        eecho "Share to be mounted must be of the form 'account.$2.core.windows.net:/account/container'."
```
sharimsft, line 228: azprefix

src/mountscript.sh
```diff
@@ -217,13 +225,47 @@ get_dir_from_share()
...
 
+get_version_from_MOUNT_OPTIONS()
+{
+    local mount_options="$MOUNT_OPTIONS"
+    local ver_string="vers="
+    local minor_ver_string="minorversion"
+    local nfs_ver=""
+    local nfs_minorvers=""
+
+    if [ -z "$mount_options" ] || [[ ! "$mount_options" == *"$ver_string"* ]]; then
```
zimchen-msft, line 243: Please confirm the previous v3 changes, and make sure no option means using v3. Once confirmed, add a comment for this if check.

src/mountscript.sh
```diff
@@ -215,15 +224,56 @@ get_dir_from_share()
     IFS=: read _ share <<< "$hostshare"
     IFS=/ read _ account container extra <<< "$share"
 
-    if [ -z "$account" -o -z "$container" -o -n "$extra" ]; then
+    if [ -z "$account" -o -z "$container" ]; then
         eecho "Bad share name: ${hostshare}."
-        eecho "Share to be mounted must be of the form 'account.blob.core.windows.net:/account/container'."
+        eecho "Share to be mounted must be of the form 'account.$2.core.windows.net:/account/container'."
```
sharimsft, line 229: $azprefix

src/mountscript.sh
```diff
@@ -215,15 +224,56 @@ get_dir_from_share()
...
+    local nfs_ver=""
+    local nfs_minorvers=""
+
+    #
+    # v3 team adds vers=3 option if version is missing in mount command.
+    # Hence, to avoid any breaking changes for v3,
+    # defaulting vers=3 if the version is not given.
+    #
+    if [ -z "$mount_options" ] || [[ ! "$mount_options" == *"$ver_string"* ]]; then
+        pecho "Adding default vers=3 mount option!"
```
sharimsft, line 250: 4.1? and below

src/mountscript.sh
```diff
@@ -602,6 +652,329 @@ ensure_aznfswatchdog()
...
+        # get the next available port
+        available_port=$(get_next_available_port)
+        if [ $? -ne 0 ]; then
+            eecho "Failed to get the next available port for nfsv4 mount."
+            return 1
+        fi
+        vecho "Next Available Port: '$available_port'"
+
+        if [ -z "$available_port" ]; then
+            eecho "Running out of ports. For Nfsv4, stunnel uses port range from $NFSV4_PORT_RANGE_START to $NFSV4_PORT_RANGE_END. All ports from this range are used by other processes."
```
sharimsft, line 668: 4.1

src/mountscript.sh
```diff
@@ -602,6 +652,329 @@ ensure_aznfswatchdog()
...
+    fi
+
+    echo "verifyChain = yes" >> $stunnel_conf_file
+    if [ $? -ne 0 ]; then
+        chattr -f +i $stunnel_conf_file
+        eecho "Failed to add verifyChain option to $stunnel_conf_file!"
+        return 1
+    fi
+
+    # TODO: checkHost value could be different for prod tenants.
```
zimchen-msft, line 764: It will for sure different for prod tenants, and we cannot tell if the mount helper is called from preprod or prod. Also depending on how the certificate is created by certificate team, this value maybe different between preprod and prod. that's the reason I said we may need an extra flag to turn off this check, but we don't have to deal it right now here.
please do verify this changes works for devfabric.

zimchen-msft (approved):

src/mountscript.sh
```diff
@@ -602,6 +652,329 @@ ensure_aznfswatchdog()
...
+    vecho "Available Port: $available_port"
+
+    if [ -z "$available_port" ]; then
+        eecho "Running out of ports. Nfsv4 has port range $NFSV4_PORT_RANGE_START to $NFSV4_PORT_RANGE_END. All ports from this range are used by other processes."
+        exit 1
+    fi
+
+    EntryExistinMountMap="true"
+
+    stunnel_conf_file="$STUNNELDIR/stunnel_$storageaccount.conf"
```
sharimsft, line 860: can be a constant

src/mountscript.sh
```diff
@@ -610,59 +983,127 @@ if ! ensure_aznfswatchdog; then
...
+            eecho "Mount failed!"
+            exit 1
+        fi
+    else
+        vecho "Mount nfs share with TLS."
+
+        IFS=/ read _ account container extra <<< "$nfs_dir"
+        storageaccount="$account"
+
+        tls_nfsv4_files_share_mount
```
sharimsft, line 1075: should you check the error code here ?

src/mountscript.sh
```diff
@@ -602,6 +652,329 @@ ensure_aznfswatchdog()
...
+
+        if [ $add_stunnel_configuration_status -ne 0 ]; then
+            eecho "Failed to add stunnel configuration to $stunnel_conf_file!"
+            chattr -i -f $stunnel_conf_file
+            rm $stunnel_conf_file
+            exit 1
+        fi
+
+        # start the stunnel process
+        stunnel_status=$(stunnel $stunnel_conf_file 2>&1)
```
sharimsft, line 889: name of the process can be a constant?

sharimsft (approved):

zimchen-msft (approved):

src/mountscript.sh
rmirashi-msft: Mount helper script first checks if the certificate already exists. If not, then it downloads the certificate.

src/mountscript.sh
rmirashi-msft: made it global

src/mountscript.sh
rmirashi-msft: $2 is $azprefix. it's the same

src/mountscript.sh
rmirashi-msft: Discussed with Raymond offline. v3 adds vers=3 option in mount command if no version is given. Hence defaulting vers=3 here to avoid breaking changes for v3.

src/mountscript.sh
```diff
@@ -215,15 +224,56 @@ get_dir_from_share()
     IFS=: read _ share <<< "$hostshare"
     IFS=/ read _ account container extra <<< "$share"
 
-    if [ -z "$account" -o -z "$container" -o -n "$extra" ]; then
```
zimchen-msft, line 218: Just realize v3 helper don't want to do sub-mount, you need to have different logic for v3 and v4

src/mountscript.sh
```diff
@@ -610,6 +970,103 @@ if ! ensure_aznfswatchdog; then
...
+
+nfs_vers=$(get_version_from_MOUNT_OPTIONS "$MOUNT_OPTIONS")
+if [ $? -ne 0 ]; then
+    echo "$nfs_vers"
+    exit 1
+fi
+
+if [ $nfs_vers == "4.1" ]; then
+    AZ_PREFIX="file"
+else
```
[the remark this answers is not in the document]
rmirashi-msft: If there is no version in mount command, v3 and v4 add default version 3. If customer adds incorrect version in mount command, v3 team fixes the version and sets the vers=3. So if the version is not 4.1, it's anyway 3. So no need of elif for vers=3 and put other unnecessary code in else block.

src/mountscript.sh
```diff
@@ -610,6 +970,103 @@ if ! ensure_aznfswatchdog; then
...
+    AZ_PREFIX="blob"
+fi
+
+nfs_host=$(get_host_from_share "$1" "$AZ_PREFIX")
+if [ $? -ne 0 ]; then
+    echo "$nfs_host"
+    exit 1
+fi
+
+# TODO: Uncomment below code later. We don't have server side changes yet, so cannot test this mount helper script on storageaccount/test tenants for v4. Hence, 'is_valid_blob_fqdn' will fail.
```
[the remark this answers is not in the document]
rmirashi-msft: modified code and added TODO comment

src/mountscript.sh
```diff
@@ -602,6 +644,324 @@ ensure_aznfswatchdog()
...
+
+#
+# Mount nfsv4 files share with TLS encryption.
+#
+tls_nfsv4_files_share_mount()
+{
+    vecho "nfs_dir=[$nfs_dir], mount_point=[$mount_point], options=[$OPTIONS], mount_options=[$MOUNT_OPTIONS]."
+
+    # Note the available port
+    available_port=$(get_next_available_port)
```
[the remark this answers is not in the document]
rmirashi-msft: if server keeps disconnecting the connection, the ports will be free and hence won't exhaust available ports.

src/mountscript.sh
```diff
@@ -602,6 +644,324 @@ ensure_aznfswatchdog()
...
+        return 1
+    fi
+
+    echo "connect = $nfs_host:$CONNECT_PORT" >> $stunnel_conf_file
+    if [ $? -ne 0 ]; then
+        chattr -f +i $stunnel_conf_file
+        eecho "Failed to add 'connect' info to $stunnel_conf_file!"
+        return 1
+    fi
+
```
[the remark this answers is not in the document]
rmirashi-msft: I put checkHost part on stash as my previous mount helper pr was messed up. And forgot to add stashed changes into this pr. But now added checkHost option in the stunnel configuration.
The checkHost value that I have used now works for devfabric and may work for preprod tenants. But for prod tenants, yes, we need to have a different checkHost value.
Instead of adding a flag, we can switch the checkHost value for preprod and prod tenants by checking $nfs_host value.
$nfs_host will have 'preprod' string for preprod tenants, so by parsing $nfs_host, we can check if its preprod or prod server and then switch the value of checkHost accordingly.

src/mountscript.sh
```diff
@@ -60,11 +68,11 @@ PID=""
...
 
 #
-# Check if the given string is a valid blob FQDN (<accountname>.blob.core.windows.net).
+# Check if the given string is a valid blob/file FQDN (<accountname>.<blob/file>.core.windows.net).
 #
-is_valid_blob_fqdn()
+is_valid_fqdn()
 {
-    [[ $1 =~ ^([a-z0-9]{3,24})(.z[0-9]+)?(.privatelink)?.blob(.preprod)?.core.(windows.net|usgovcloudapi.net|chinacloudapi.cn)$ ]]
+    [[ $1 =~ ^([a-z0-9]{3,24})(.z[0-9]+)?(.privatelink)?.$2(.preprod)?.core.(windows.net|usgovcloudapi.net|chinacloudapi.cn)$ ]]
```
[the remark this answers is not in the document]
rmirashi-msft: I am using $1. I am parsing $1 here

src/mountscript.sh
rmirashi-msft: $2 = $azprefix. Its the same

src/mountscript.sh
rmirashi-msft: I replied to your comment from previous iteration and explained how we can switch the value of checkHost for preprod and prod.
Also, I verified that this checHost works for devfabric.

src/mountscript.sh
rmirashi-msft: no, cannot be a constant. We are changing the attributes of stunnel_conf_file later using chattr command that needs write permission, so cannot make it constant.

src/mountscript.sh
rmirashi-msft: not required. We anyway exit after calling this function for v4. Also, in this function, if anything fails, we exit from there so there is no need to check the error code here.

src/mountscript.sh
rmirashi-msft: I looked into the code. We are handling the errors slightly differently at all the places. So even if I write a wrapper function to start the stunnel process and handle the errors, the callers will still have to parse the status code and figure out why stunnel failed to start and do the handling. So in my opinion, we won't save much lines by having a common function to start stunnel process.

src/mountscript.sh
rmirashi-msft: handled it

zimchen-msft (approved):

PR #93, opened 2023-12-11
Adding CHANGELOG.md file and print AZNFS version for debugging in aznfswatchdog

palashvij-msft:


README.md
```diff
@@ -40,14 +40,26 @@ AZNFS is supported on following Linux distros:
 	```
 	It will install the aznfs mount helper program and the aznfswatchdog service.
 
+## Auto Update AZNFS
+
+- Upon running the installation command, you will be prompted to configure automatic updates for AZNFS. Enabling automatic updates ensures that you 
+  stay current with the latest features, improvements, and security patches, providing you with the best and most seamless AZNFS experience. 
+  You can turn off/on auto-update at any time from /opt/microsoft/aznfs/data/config and update respectively.
```
linuxsmiths, line 47: You can also turn off/on auto-update at any time by changing the value of AUTO_UPDATE_AZNFS variable in /opt/microsoft/aznfs/data/config.

There seems to be a repetition of this, so just say once.

src/aznfswatchdog
```diff
@@ -307,13 +307,49 @@ run_daily_once()
...
+    fi
+
+    echo "$distro_lower"
+}
+
+detect_aznfs_version()
+{
+    local distro="$distro_id"
+
+    sleep 45
```
linuxsmiths, line 330: why?

src/aznfswatchdog
```diff
@@ -307,13 +307,49 @@ run_daily_once()
...
+        current_version=$(zypper info aznfs_sles 2>/dev/null | grep "^Version" | tr -d " " | cut -d ':' -f2 | cut -d '-' -f1)
+    fi
+
+    vecho "AZNFS version: $current_version"
+}
+
 vecho "Starting aznfswatchdog..."
 
-# Detect and log distro and bash version
+# Detect and log distro, bash and AZNFS-mount version
```
linuxsmiths, line 345: move L346 to 361 into a single function and call it log_version_info()
don't need the detect_aznfs_version() function and also don't do it in the background.

src/aznfswatchdog
palashvij-msft: example for sles, it takes few seconds after which we are able to fetch the right version number, otherwise it is blank output. For similar reason only we deferred auto-update to after 60 seconds to starting watchdog. So, I made this a function a background process.

src/aznfswatchdog
palashvij-msft: background as it takes few seconds before we are able to fetch the aznfs version after starting watchdog, specifically for sles.

src/aznfswatchdog
linuxsmiths: let's discuss

linuxsmiths (approved):

linuxsmiths (approved):

PR #98, opened 2023-12-18
update variable to avoid mix-up with update mountmap

palashvij-msft:


linuxsmiths (approved):

src/mountscript.sh
```diff
@@ -602,6 +602,28 @@ ensure_aznfswatchdog()
...
 
+#
+# To maintain consistency in case of regional account and in general to avoid creating
+# multiple DNAT entries corrosponding to one LOCAL_IP, first check for resolved IP in mountmap.
+# This will help keep mountmap and DNAT entries in sync with each other.
+#
+resolve_ipv4_with_preference_to_mountmap()
+{
+        local fqdn=$1
+        local mountmap_entry=$(grep -m1 "^${fqdn} " $MOUNTMAP)
```
linuxsmiths, line 613: we should hold a lock on MOUNTMAP to avoid reading garbled entry if aznfswatchdog is changing it.
This ensures that we either read the mountmap entry before aznfswatchdog changes it (and the dnat rule) or after it, in both cases we can use the entry with the confidence that aznfswatchdog will properly handle the IP address change for this fqdn after this point.

src/mountscript.sh
```diff
@@ -602,6 +602,28 @@ ensure_aznfswatchdog()
...
+
+        IFS=" " read _ local_ip old_nfs_ip <<< "$mountmap_entry"
+        if [ -n "$old_nfs_ip" ]; then
+            echo $old_nfs_ip
```
linuxsmiths, line 617: nit "$old_nfs_ip"

src/mountscript.sh
```diff
@@ -637,12 +659,17 @@ if ! is_valid_blob_fqdn "$nfs_host"; then
...
-if [ $? -ne 0 ]; then
-    echo "$nfs_ip"
-    eecho "Cannot resolve IP address for ${nfs_host}!"
-    eecho "Mount failed!"
-    exit 1
+nfs_ip=$(resolve_ipv4_with_preference_to_mountmap "$nfs_host")
+status=$?
+if [ $status -ne 0 ]; then
+    if [ $status -eq 2 ]; then
+        vecho "Resolving the IP address for FQDN: $nfs_host to $nfs_ip from mountmap"
```
linuxsmiths, line 666: "Resolved IP address for FQDN from mountmap [$nfs_host -> $nfs_ip]"

src/aznfswatchdog
```diff
@@ -607,10 +607,10 @@ while :; do
             if ! update_mountmap_entry "$line" "$l_host $l_ip $new_ip"; then
                 eecho "Will reattempt the operation in next iteration."
             else
-                mountpoint=$(echo "$findmnt" | grep " nfs ${l_ip}:" | awk '{print $4}')
-                mountpoint=$(echo -e "$mountpoint")
-                if [ -n "$mountpoint" ]; then
-                    ping_new_endpoint "$mountpoint" &
+                mountpoint2048=$(echo "$findmnt" | egrep " nfs ${l_ip}:.*\<port=2048\>" | awk '{print $4}' | head -n 1)
```
linuxsmiths, line 610: you can omit the head -n1 by using -m1 option to egrep

mountpoint2048=$(echo "$findmnt" | egrep -m1 " nfs ${l_ip}:.*\<port=2048\>" | awk '{print $4}')

src/mountscript.sh
```diff
@@ -602,6 +602,28 @@ ensure_aznfswatchdog()
...
+        IFS=" " read _ local_ip old_nfs_ip <<< "$mountmap_entry"
+        if [ -n "$old_nfs_ip" ]; then
+            echo $old_nfs_ip
+            return 2 
+        fi
+
+        #
+        # Resolve FQDN to IPv4 using DNS if not found in the mountmap.
+        #
+        resolve_ipv4 "$nfs_host" "true"
```
linuxsmiths, line 624: nit, use local variable $fqdn in place of the global variable $nfs_host

linuxsmiths (approved):

src/aznfswatchdog
```diff
@@ -133,24 +133,29 @@ check_for_regional_account()
...
+    local seconds_remaining=$5
 
     key="${l_ip}:${l_sport}:${l_dport}:${l_nfsip}"
-    let cthash_synsent[$key]++
-    let cthash_synsent_now[$key]++
+    
+    if [ ! -v cthash_synsent[$key] ]; then
+        cthash_synsent[$key]=$seconds_remaining
+        return
+    fi
```
linuxsmiths, line 150: comment above this

`#` First time we are seeing this conntrack entry.

src/aznfswatchdog
```diff
@@ -133,24 +133,29 @@ check_for_regional_account()
...
+
+    age_seconds=$(expr ${cthash_synsent[$key]} - $seconds_remaining)
```
linuxsmiths, line 152: `#`
`#` How long has this entry been around?
`#` If it's around for more than 25-30 secs, we consider the entry as "stuck" and delete it to cause fresh entry to
`#` be created, and help make progress.
`#`

you can remove the comment block which is now incorrect and move this comment block above the age_seconds= line.

src/aznfswatchdog
```diff
@@ -133,24 +133,29 @@ check_for_regional_account()
...
 
     #
     # We are called every 5 secs, so this deletes an entry stuck in
     # SYN_SENT for 25/30 secs.
     #
-    if [ ${cthash_synsent[$key]} -ge 5 ]; then
+    if [ $age_seconds -ge 25 ]; then
         cmd="conntrack -D -p tcp -d $l_ip -r $l_nfsip --sport $l_sport --dport $l_dport"
         wecho "Deleting conntrack entry stuck in SYN_SENT state [$cmd]"
```
linuxsmiths, line 160: "Deleting conntrack entry stuck in SYN_SENT state for $age_seconds seconds [$cmd]"

src/aznfswatchdog
```diff
@@ -170,24 +175,29 @@ reconcile_conntrack_synsent()
 # a new conntrack entry, if the IP were to change again soon.
 #
 declare -A cthash_unreplied
-declare -A cthash_unreplied_now
 
 reconcile_conntrack_unreplied()
```
linuxsmiths, line 179: add same comments as above in this function too

src/aznfswatchdog
```diff
@@ -221,14 +225,18 @@ reconcile_conntrack()
 
     if [ -n "$output" ]; then
         for entry in $output; do
-            # XXX Remove this log after running for few days.
+            # XXX Remove this log after running for few days.(not too noisy!)
```
linuxsmiths, line 228: nit, remove period after days and put it at the end of line.

src/aznfswatchdog
```diff
@@ -221,14 +225,18 @@ reconcile_conntrack()
...
             vecho "$entry"
 
-            matchstr=".* SYN_SENT src=.* dst=$l_ip sport=([0-9]+) dport=([0-9]+).*"
+            # pasting here for ease during review (remove later)
```
linuxsmiths, line 231: instead of this, say

`#` Sample conntrack entry.

and we can leave it around

src/aznfswatchdog
```diff
@@ -221,14 +225,18 @@ reconcile_conntrack()
...
+            # tcp      6 114 SYN_SENT src=10.20.0.17 dst=10.161.100.101 sport=819 dport=2048 [UNREPLIED] src=20.150.35.196 dst=10.20.0.17 sport=2048 dport=819 mark=0 use=1
+            matchstr="tcp \s*[0-9]+ \s*([0-9]+).* SYN_SENT src=.* dst=$l_ip sport=([0-9]+) dport=([0-9]+).*"
```
linuxsmiths, line 233: since regexp tries to find the longest match we have to be careful that the sport and dport that match are not in the reply direction. Also to be safe let's use \s* for space as it can be a space or tab or multiple spaces.

matchstr="tcp\s+[0-9]+\s+([0-9]+)\s+SYN_SENT\s+src=[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+\s+dst=${l_ip}\s+sport=([0-9]+)\s+dport=([0-9]+).*"

src/aznfswatchdog
```diff
@@ -221,14 +225,18 @@ reconcile_conntrack()
...
             if [[ "$entry" =~ $matchstr ]]; then
-                l_sport=${BASH_REMATCH[1]}
-                l_dport=${BASH_REMATCH[2]}
-                reconcile_conntrack_synsent $l_ip $l_sport $l_dport $l_nfsip
+                l_seconds_remaining=${BASH_REMATCH[1]}
+                l_sport=${BASH_REMATCH[2]}
+                l_dport=${BASH_REMATCH[3]}
+                vecho "l_seconds_remaining (syn_sent): $l_seconds_remaining" # remove later
```
linuxsmiths, line 238: remove this now

src/aznfswatchdog
```diff
@@ -238,34 +246,22 @@ reconcile_conntrack()
...
+            # pasting here for ease during review (remove later)
+            # tcp      6 299 ESTABLISHED src=10.2.4.4 dst=10.161.100.100 sport=1015 dport=2048 [UNREPLIED] src=20.60.236.11 dst=10.2.4.4 sport=2048 dport=1015 mark=0 use=1 
+            matchstr="tcp \s*[0-9]+ \s*([0-9]+) .* dst=$l_ip sport=([0-9]+) dport=([0-9]+) \[UNREPLIED\] src=([0-9]+\.[0-9]+\.[0-9]+\.[0-9]+).*"
             if [[ "$entry" =~ $matchstr ]]; then
-                l_sport=${BASH_REMATCH[1]}
-                l_dport=${BASH_REMATCH[2]}
-                l_reply_srcip=${BASH_REMATCH[3]}
-                reconcile_conntrack_unreplied $l_ip $l_sport $l_dport $l_reply_srcip
+                l_seconds_remaining=${BASH_REMATCH[1]}
+                l_sport=${BASH_REMATCH[2]}
```
linuxsmiths, line 257: same comments as above

linuxsmiths (commented):
LG, few comments.

src/aznfswatchdog
palashvij-msft: Just a confirmation, age_seconds is the time >25, from the first detection by our code (not exact time). since we check every 5 seconds, so I also feel, it's mostly okay to print that. At max 5 seconds of difference bw exact and caught duration.

linuxsmiths (approved):

PR #103, opened 2024-01-12
BASIC setup for testing AZNFS

palashvij-msft:


.github/workflows/aznfs-test.yml
```diff
@@ -0,0 +1,354 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      versionName:
+        description: 'Version Name'
+        required: true
+      storageAccounts:
+        description: 'Space-separated Storage Accounts'
+        required: true
+
```
linuxsmiths, line 12: can we define an env block here and put all env variables needed by any step in that. f.e. I can see at least following coming over there

RELEASE_NUMBER
STORAGE_ACCOUNTS

having all env variables in one place is more readable

.github/workflows/aznfs-test.yml
```diff
@@ -0,0 +1,354 @@
...
+    name: Test AZNFS - Ubuntu 20
+    runs-on: self-hosted-ubuntu20
+    needs: test_aznfs_ubuntu18
+    steps:
+      - name: Install Prerequisites
+        shell: bash
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y build-essential
+      - name: Check Version and Storage Account
```
linuxsmiths, line 89: can we create a composite action (https://docs.github.com/en/actions/creating-actions/creating-a-composite-action) called prep-aznfs-test-machine which runs a shell script that correctly preps the machine on which it runs. 
Then we just call that action from here followed by the actual test call.
This will reduce code repetition.

PR #116, opened 2024-05-13
Fingerprinting AZNFS changes

palashvij-msft:


src/mountscript.sh
```diff
@@ -773,6 +776,31 @@ if [ -z "$AZNFS_PMAP_PROBE" -o "$AZNFS_PMAP_PROBE" == "0" ]; then
...
+# Perform a pseudo mount to generate a gatepass for the actual mount call.
+# This request is expected to fail with "server access denied" if server-side changes are enabled,
+# or with "no such file or directory" if not. Failure of this call is expected behavior, 
+# and we proceed normally when it occurs.
+#
+mount_output=$(mount -t nfs $OPTIONS -o "$MOUNT_OPTIONS" "${LOCAL_IP}:${nfs_dir}/$AZNFS_FINGERPRINT" "$mount_point" 2>&1)
+mount_status=$?
+
+if [ -n "$mount_output" ]; then
+    vecho "$mount_output"
```
linuxsmiths, line 789: nit, for clarity, let's do

vecho "[Gatepass mount] $mount_output"

src/mountscript.sh
```diff
@@ -773,6 +776,31 @@ if [ -z "$AZNFS_PMAP_PROBE" -o "$AZNFS_PMAP_PROBE" == "0" ]; then
...
+# Check if the mount operation failed (expected behavior).
+#
+if [ $mount_status -ne 0 ]; then
+    vecho "Gatepass mount failed, as expected!"
+else
+    #
+    # Exit with an error code if the mount operation succeeded, which is unexpected.
+    #
+    exit 1
+fi
```
linuxsmiths, line 802: no need to log again in case of success, as we already logged above, just say

`
#
# Ensure that gatepass mount operation failed (expected behavior).
# Exit with an error code if it succeeded, which is unexpected.
#
if [ $mount_status -eq 0 ]; then
    eecho "[Gatepass mount] Unexpected success!"
    exit 1
fi
`

linuxsmiths (approved):

PR #119, opened 2024-06-03
Removing Extra Mount Logs from Console

cagrawal-msft:


src/mountscript.sh
```diff
@@ -83,6 +93,12 @@ check_nconnect()
     if [[ "$MOUNT_OPTIONS" =~ $matchstr ]]; then
         value="${BASH_REMATCH[1]}"
         if [ $value -gt 1 ]; then
+            if [ $USING_PORT_2047 == true ]; then
+                pecho "No need to use nconnect with port 2047, setting nconnect=0!"
+                MOUNT_OPTIONS=$(echo "$MOUNT_OPTIONS" | sed "s/\<nconnect\>=$value/nconnect=0/g")
+                return 0
+            fi
+
```
linuxsmiths, line 101: this is not correct, we need nconnect= option for port 2047.
also L116-119 is forcing nconnect=4 for both 2048 and 2047 mounts, we need it only for 2047
also, we cannot mix 2048 and 2047 nconnect mounts, as 2048 needs the azure nconnect enabled while 2047 wants it disabled. 
also, vvecho at L97 and L117
Following function takes care of all these

```
check_nconnect()
{
    matchstr="\<nconnect\>=([0-9]+)"
    if [[ "$MOUNT_OPTIONS" =~ $matchstr ]]; then
        value="${BASH_REMATCH[1]}"
        if [ $value -gt 1 ]; then
            # Load sunrpc module if not already loaded.
            if [ ! -d /sys/module/sunrpc/ ]; then
                modprobe sunrpc
            fi

            #
            # W/o server side nconnect, we need the azure nconnect support,
            # turn it on. OTOH, if Server side nconnect is being used turn off
            # azure nconnect support if enabled.
            #
            if [ $USING_PORT_2047 == false ]; then
                if has_2047_nconnect_mounts; then
                    eecho "One or more mounts to port 2047 are using nconnect."
                    eecho "Cannot mix port 2048 and 2047 nconnect mounts, unmount those and try mounting again!"
                    return 1
                fi

                if [ ! -e /sys/module/sunrpc/parameters/enable_azure_nconnect ]; then
                    eecho "nconnect option needs NFS client with Azure nconnect support!"
                    return 1
                fi

                # Supported, enable if not enabled.
                enabled=$(cat /sys/module/sunrpc/parameters/enable_azure_nconnect)
                if ! [[ "$enabled" =~ [yY] ]]; then
                    vvecho "Azure nconnect not enabled, enabling!"
                    echo Y > /sys/module/sunrpc/parameters/enable_azure_nconnect
                fi
            else
                if has_2048_nconnect_mounts; then
                    eecho "One or more mounts to port 2048 are using nconnect."
                    eecho "Cannot mix port 2048 and 2047 nconnect mounts, unmount those and try mounting again!"
                    return 1
                fi

                if [ -e /sys/module/sunrpc/parameters/enable_azure_nconnect ]; then
                    enabled=$(cat /sys/module/sunrpc/parameters/enable_azure_nconnect)
                    if [[ "$enabled" =~ [yY] ]]; then
                        vvecho "Azure nconnect enabled, disabling!"
                        echo N > /sys/module/sunrpc/parameters/enable_azure_nconnect
                    fi
                fi

                #
                # Higher nconnect values don't work well for server side
                # nconnect, limit to optimal value 4.
                #
                if [ $value -gt 4 ]; then
                    vvecho "Suboptimal nconnect value $value, forcing nconnect=4!"
                    MOUNT_OPTIONS=$(echo "$MOUNT_OPTIONS" | sed "s/\<nconnect\>=$value/nconnect=4/g")
                fi
            fi
        fi
    fi
}
```

has_204[78]_nconnect_mounts() can be defined above as

```

#
# Check if any nconnect mount exists for port 2048.
#
has_2048_nconnect_mounts()
{
    local findmnt=$(findmnt --raw --noheading -o MAJ:MIN,FSTYPE,SOURCE,TARGET,OPTIONS -t nfs | egrep "\<port=2048\>" | egrep "\<nconnect=")

    [ -n "$findmnt" ]
}

#
# Check if any nconnect mount exists for port 2047.
#
has_2047_nconnect_mounts()
{
    local findmnt=$(findmnt --raw --noheading -o MAJ:MIN,FSTYPE,SOURCE,TARGET,OPTIONS -t nfs | egrep "\<port=2047\>" | egrep "\<nconnect=")

    [ -n "$findmnt" ]
}
```

src/mountscript.sh
```diff
@@ -653,6 +682,9 @@ parse_arguments()
         else
             OPTIONS="$OPTIONS $arg"
         fi
+
+        if [ "$arg" == "-v" || "$arg" == "--verbose" ]
+            VERBOSE_ENABLED=true
```
linuxsmiths, line 687: nit, move it to L684, where it makes more sense

PR #136, opened 2024-07-23
resolve minor number from combined device id retrieved from stat call

palashvij-msft:
EXAMPLE LOG WITH DEBUG ADDED TO VERIFY IF WORKING CORRECTLY-

Mounting container529 on /mnt/palash529...
Adding nolock mount option!
Adding retrans=6 mount option!
mount.nfs: timeout set for Tue Jul 23 06:18:00 2024
mount.nfs: trying text-based options 'vers=3,proto=tcp,sec=sys,nolock,retrans=6,port=2048,mountport=2048,addr=10.161.100.105'
Mount completed: palashvijstorageaccount.blob.core.windows.net:/palashvijstorageaccount/container529 on /mnt/palash529 using proxy IP 10.161.100.105 and endpoint IP 20.60.185.161
Read ahead size for /mnt/palash529 set to 16384 KB! **_[debug]BLOCK ID=2097225, MINOR=585_**
Mounting container530 on /mnt/palash530...
Adding nolock mount option!
Adding retrans=6 mount option!
mount.nfs: timeout set for Tue Jul 23 06:18:00 2024
mount.nfs: trying text-based options 'vers=3,proto=tcp,sec=sys,nolock,retrans=6,port=2048,mountport=2048,addr=10.161.100.105'
Mount completed: palashvijstorageaccount.blob.core.windows.net:/palashvijstorageaccount/container530 on /mnt/palash530 using proxy IP 10.161.100.105 and endpoint IP 20.60.185.161
Read ahead size for /mnt/palash530 set to 16384 KB! [**_debug]BLOCK ID=2097226, MINOR=586_**
Mounting container531 on /mnt/palash531...
Adding nolock mount option!
Adding retrans=6 mount option!

PR #137, opened 2024-07-23
resolve minor number from combined device id retrieved from stat call

palashvij-msft:
resolve minor number from combined device id retrieved from stat call… (#136)

* minor from combined device id

* addressing PR comments and updating changelog


linuxsmiths (approved):
Thanks!

PR #143, opened 2024-09-18
Personal/sroghanchi/fix nfsv4 bugs

sroghanchi:
Fixing mount helper bugs and race conditions.


src/nfsv4mountscript.sh
```diff
@@ -24,14 +24,22 @@ CERT_PATH=
 CERT_UPDATE_COMMAND=
 STUNNEL_CAFILE=
 
+# Temporary mountmap file used to overwrite mountmap file.
+TMP_MOUNTMAPv4="/tmp/mountmapv4.tmp"
```
sharimsft, line 28: are you sure they will always have this /tmp folder?

src/nfsv4mountscript.sh
```diff
@@ -125,6 +133,28 @@ install_CA_cert()
...
 
+get_check_host_value()
+{
+    local hostname=$1
+    local check_host_value="*.file.core.windows.net"
+
+    declare -A certs
+    certs=(
+        ["preprod.core.windows.net$"]="*.file.preprod.core.windows.net"
+        ["chinacloudapi.cn$"]="*.file.core.usgovcloudapi.net"
```
sharimsft, line 144: should this be in a configuration file?

src/aznfswatchdogv4
```diff
@@ -199,15 +245,44 @@ process_nfsv4_mounts()
...
+                exec {fd1}<$l_conf
+                flock -e $fd1
+                exec {fd2}<$MOUNTMAPv4
+                flock -e $fd2
 
                 # Delete IFF mountmap is not changed since we read it above.
                 l_mtime=$(ensure_mountmapv4_not_exist "$line" "$mtime_mountmap")
-                [ $? -eq 0 ] && mtime_mountmap=$l_mtime
+
+                #
```
sharimsft, line 257: duplicated code?

src/nfsv4mountscript.sh
```diff
@@ -267,6 +296,48 @@ tls_nfsv4_files_share_mount()
...
+                    sed "\#$stunnel_conf_file;#s#\(;mounted\|;failed\)#;waiting#" $MOUNTMAPv4 > $TMP_MOUNTMAPv4
+                    # Add mount timeout to the mountmap entry. Used when aznfsWatchdog is cleaning up the mountmap file.
+                    sed -i "\#$stunnel_conf_file;#s#\(.*;\)\([^;]*\)\$#\1$mount_timeout#" $TMP_MOUNTMAPv4
+                    cp $TMP_MOUNTMAPv4 $MOUNTMAPv4
+                    rm $TMP_MOUNTMAPv4
+                    chattr -f +i $MOUNTMAPv4
+                else
+                    # We should always have the mountmap entry for the stunnel_conf_file.
+                    # TODO: Double check if we need to handle this case.
+                    eecho "Failed to find the mountmap entry for $stunnel_conf_file in $MOUNTMAPv4."
```
sharimsft, line 328: unlock here?

src/nfsv4mountscript.sh
```diff
@@ -361,23 +445,48 @@ tls_nfsv4_files_share_mount()
...
+        chattr -f +i $MOUNTMAPv4
+        flock -u $fd2
+        exec {fd2}<&-
         eecho "Mount failed!"
         exit 1
+    else
+        vecho "Updating mountmap status to mounted."
+        sed "\#$stunnel_conf_file;#s#;waiting#;mounted#" $MOUNTMAPv4 > $TMP_MOUNTMAPv4
+        cp $TMP_MOUNTMAPv4 $MOUNTMAPv4
+        rm $TMP_MOUNTMAPv4
```
sharimsft, line 485: repeated code. can it be moved out?

sharimsft (changes requested):
i have done some review as requested

src/nfsv4mountscript.sh
sroghanchi: Yes, but I can change it to concatenate the moutnmap instead of using the tmp file.

lib/common.sh
```diff
@@ -683,9 +686,37 @@ verify_iptable_entry()
     fi
 }
 
+# Find CheckHost value for stunnel configuration based on storage account hostname.
+get_check_host_value()
+{
+    local hostname=$1
```
palashvij-msft, line 692: Validate hostname parameter? It is not empty?

lib/common.sh
```diff
@@ -683,9 +686,37 @@ verify_iptable_entry()
...
+    local check_host_value="*.file.core.windows.net"
+
+    declare -A certs
+    certs=(
+        ["preprod.core.windows.net$"]="*.file.preprod.core.windows.net"
+        ["chinacloudapi.cn$"]="*.file.core.usgovcloudapi.net"
```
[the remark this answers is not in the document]
sroghanchi: I addressed it. The feature is only enabled in preprod so only the preprod cert is tested so far. We will test the rest later as we enable the feature.

scripts/aznfs_install.sh
```diff
@@ -132,8 +132,8 @@ perform_aznfs_update()
         AZNFS_RELEASE_SUSE="aznfs_sles-${RELEASE_NUMBER}-1"
         package_name=${AZNFS_RELEASE_SUSE}.x86_64.rpm
     else
-        if [ "$distro_id" == "mariner" ]; then
-            AZNFS_RELEASE="aznfs_mariner-${RELEASE_NUMBER}-1"
+        if [[ "$distro_id" == "mariner" || (("$distro_id" == "rhel" || "$distro_id" == "centos") && $(grep '^VERSION_ID=' /etc/os-release | cut -d'=' -f2 | tr -d '"' | cut -d'.' -f1) == 7) ]] ; then
```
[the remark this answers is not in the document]
sroghanchi: Stunnel version that comes with Centos7 and RedHat7 are old and not compatible with the version needed by mount helper, that's why we should install it from the source code. @palashvij-msft Please let me know if you have any naming suggestions, if not I can comment on this in README.

scripts/aznfs_install.sh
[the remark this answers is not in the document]
sroghanchi: It was added by Richa with the initial NFSv4 design, I've removed it.

lib/common.sh
```diff
@@ -683,9 +686,37 @@ verify_iptable_entry()
...
+    )
+
+    for cert in "${!certs[@]}"; do
+        if [[ "$hostname" =~ $cert ]]; then
+                check_host_value="${certs[$cert]}"
+                break
+        fi
+    done
+
+    echo $check_host_value
```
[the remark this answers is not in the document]
sroghanchi: Addressed.

lib/common.sh
```diff
@@ -31,6 +31,9 @@ HOSTNAME=$(hostname)
 
 LOCALHOST="127.0.0.1"
 
+# Command to use for getting socket statistics - netstat or ss
```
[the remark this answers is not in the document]
sroghanchi: Addressed.

lib/common.sh
```diff
@@ -683,9 +686,37 @@ verify_iptable_entry()
...
+}
+
 # On some distros mount program doesn't pass correct PATH variable.
 export PATH=$PATH:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 
+if command -v netstat &> /dev/null; then
+    NETSTATCOMMAND="netstat"
+elif command -v ss &> /dev/null; then
+    NETSTATCOMMAND="ss"
+fi
```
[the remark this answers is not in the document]
sroghanchi: Addressed.

src/mountscript.sh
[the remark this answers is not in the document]
sroghanchi: I just rebased it.

packaging/aznfs/DEBIAN/postinst
```diff
@@ -12,6 +12,61 @@ FLAG_FILE="/tmp/.update_in_progress_from_watchdog.flag"
 CONFIG_FILE="/opt/microsoft/aznfs/data/config"
 AUTO_UPDATE_AZNFS="false"
 
+install_stunnel()
+{
+    # Install stunnel package from source
```
palashvij-msft, line 17: We can some kind of check for dependencies here to be safe (since I see make is part of Recommends and not Requires)-

    for cmd in wget tar make gcc; do
        command -v $cmd >/dev/null || { echo "$cmd is required but not installed. Exiting."; exit 1; }
    done

package.sh
```diff
@@ -196,7 +181,7 @@ dpkg-deb -Zgzip --root-owner-group --build $STG_DIR/deb/$pkg_dir
 
 generate_rpm_package rpm
 generate_rpm_package suse
-generate_rpm_package mariner
+generate_rpm_package stunnel
```
palashvij-msft, line 184: Can you clarify why this needs to be separate? Is it being used explicitly somewhere? How does "generate_rpm_package rpm" differ from "generate_rpm_package stunnel"?

packaging/aznfs/RPM/aznfs.spec
```diff
@@ -4,11 +4,8 @@ Release: 1
 Summary: Mount helper program for correctly handling endpoint IP address changes for Azure Blob NFS mounts and providing a secure communication channel for Azure File NFS mounts
 License: MIT
 URL: https://github.com/Azure/AZNFS-mount/blob/main/README.md
-%if 0%{?mariner}
-Requires: bash, PROCPS_PACKAGE_NAME, conntrack-tools, iptables, bind-utils, iproute, util-linux, nfs-utils, NETCAT_PACKAGE_NAME, newt, net-tools, build-essential, binutils, kernel-headers, openssl, openssl-devel
-%else
-Requires: bash, PROCPS_PACKAGE_NAME, conntrack-tools, iptables, bind-utils, iproute, util-linux, nfs-utils, NETCAT_PACKAGE_NAME, newt, stunnel, net-tools
-%endif
+Requires: bash, PROCPS_PACKAGE_NAME, conntrack-tools, iptables, bind-utils, iproute, util-linux, nfs-utils, NETCAT_PACKAGE_NAME, newt, net-tools, binutils, kernel-headers, openssl, openssl-devel, gcc
+Recommends: build-essential, make
```
palashvij-msft, line 8: If make is missing, stunnel building might not be possible?
Why are we choosing Recommends here? Any specific reason? AFAIK, Recommends will make installation of packages optional right?

packaging/aznfs/DEBIAN/postinst
```diff
@@ -12,6 +12,61 @@ FLAG_FILE="/tmp/.update_in_progress_from_watchdog.flag"
 CONFIG_FILE="/opt/microsoft/aznfs/data/config"
 AUTO_UPDATE_AZNFS="false"
 
+install_stunnel()
```
palashvij-msft, line 15: If an older version of stunnel is already installed, are we okay to update it to the version we find best compatible? Is there something like a check can be made to determine if the version already installed meets the requirements?

Can you help me understand this scenario?

packaging/aznfs/DEBIAN/postinst
```diff
@@ -12,6 +12,61 @@ FLAG_FILE="/tmp/.update_in_progress_from_watchdog.flag"
...
+        rm -rf /tmp/$stunnel_dir
+        rm -f /tmp/stunnel-5.73.tar.gz
+        exit 1
+    fi
+
+    make
+    if [ $? -ne 0 ]; then
+        echo "Failed to build stunnel. Please install stunnel and try again."
+        cd -
+        rm -rf /tmp/$stunnel_dir
```
palashvij-msft, line 48: these 4 commands look very redundant-

        cd -
        rm -rf /tmp/$stunnel_dir
        rm -f /tmp/stunnel-5.73.tar.gz
        exit 1
        
 can we reduce repeated thing and use a seperate function or something better?

packaging/aznfs/DEBIAN/postinst
```diff
@@ -12,6 +12,61 @@ FLAG_FILE="/tmp/.update_in_progress_from_watchdog.flag"
...
+        rm -f /tmp/stunnel-5.73.tar.gz
+        exit 1
+    fi
+    cd -
+    rm -rf /tmp/$stunnel_dir
+    rm -f /tmp/stunnel-5.73.tar.gz
+
+    sudo ln -s /usr/local/bin/stunnel /bin/stunnel
```
palashvij-msft, line 67: We may want to force here, else it might fail?
sudo ln -sf /usr/local/bin/stunnel /bin/stunnel

packaging/aznfs/DEBIAN/postinst
palashvij-msft, line 1: similar comments for .spec as well

package.sh
linuxsmiths: yes, I thought you said you don't need a separate package for stunnel now. The package will be either rpm (for non-suse distros) or suse (for suse).

sroghanchi: Yes, I removed the extra packages but forgot to cleanup this part of the script.

packaging/aznfs/DEBIAN/postinst
sroghanchi: I changed that to only install staunnel if either it's not installed or the version is older than the expected version.

packaging/aznfs/DEBIAN/postinst
sroghanchi: Done.

packaging/aznfs/DEBIAN/postinst
sroghanchi: Done.

packaging/aznfs/DEBIAN/postinst
sroghanchi: Done.

lib/common.sh
```diff
@@ -683,9 +686,40 @@ verify_iptable_entry()
...
 export PATH=$PATH:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 
+if command -v netstat &> /dev/null; then
+    NETSTATCOMMAND="netstat"
+elif command -v ss &> /dev/null; then
+    NETSTATCOMMAND="ss"
+else
+    eecho "[FATAL] Neither 'netstat' nor 'ss' is available!"
+    exit 1
+fi
```
linuxsmiths, line 722: since v3 is not using it, let's not make this a fatal error. v4 can check for it and bail out if it's not set.

packaging/aznfs/DEBIAN/postinst
```diff
@@ -11,6 +11,62 @@ set -e
...
 CONFIG_FILE="/opt/microsoft/aznfs/data/config"
 AUTO_UPDATE_AZNFS="false"
+INSTALLED_STUNNEL_VERSION="0"
+REQUIRED_STUNNEL_VERSION="5.73"
+
+cleanup_stunnel_files()
+{
+    cd -
+    rm -rf /tmp/$stunnel_dir
+    rm -f /tmp/stunnel-5.73.tar.gz
```
linuxsmiths, line 21: replace all uses of hardcoded 5.73 with $REQUIRED_STUNNEL_VERSION

packaging/aznfs/DEBIAN/postinst
```diff
@@ -11,6 +11,62 @@ set -e
...
+    if [ $? -ne 0 ]; then
+        echo "Failed to download stunnel source code. Please install stunnel and try again."
+        exit 1
+    fi
+
+    tar -xvf /tmp/stunnel-5.73.tar.gz -C /tmp
+    if [ $? -ne 0 ]; then
+        echo "Failed to extract stunnel tarball. Please install stunnel and try again."
+        cd -
+        rm -f /tmp/stunnel-5.73.tar.gz
```
linuxsmiths, line 37: we can call cleanup_stunnel_files

packaging/aznfs/DEBIAN/postinst
```diff
@@ -11,6 +11,62 @@ set -e
...
+        exit 1
+    fi
+
+    stunnel_dir=$(tar -tf /tmp/stunnel-5.73.tar.gz | head -n 1 | cut -f1 -d'/')
```
linuxsmiths, line 41: directly use stunnel version

packaging/aznfs/DEBIAN/postinst
```diff
@@ -11,6 +11,62 @@ set -e
...
+    fi
+
+    make
+    if [ $? -ne 0 ]; then
+        echo "Failed to build stunnel. Please install stunnel and try again."
+        cleanup_stunnel_files
+        exit 1
+    fi
+
+    sudo rm -f /bin/stunnel
```
linuxsmiths, line 58: move this just before the ln command

packaging/aznfs/DEBIAN/postinst
```diff
@@ -11,6 +11,62 @@ set -e
...
+
+    make install
+    if [ $? -ne 0 ]; then
+        echo "Failed to install stunnel. Please install stunnel and try again."
+        cleanup_stunnel_files
+        exit 1
+    fi
+
+    cleanup_stunnel_files
+    sudo ln -sf /usr/local/bin/stunnel /bin/stunnel
```
linuxsmiths, line 68: this is not needed

packaging/aznfs/DEBIAN/postinst
```diff
@@ -11,6 +11,62 @@ set -e
...
+    fi
+
+    sudo rm -f /bin/stunnel
+
+    make install
+    if [ $? -ne 0 ]; then
+        echo "Failed to install stunnel. Please install stunnel and try again."
+        cleanup_stunnel_files
+        exit 1
+    fi
```
linuxsmiths, line 65: in case make install fails, if we don't want to be left w/o a workign stunnel version we can replace the aboce rm with mv and then mv it back in case of failure.

packaging/aznfs/DEBIAN/postinst
```diff
@@ -11,6 +11,62 @@ set -e
...
+
+    cleanup_stunnel_files
+    sudo ln -sf /usr/local/bin/stunnel /bin/stunnel
+}
```
linuxsmiths, line 69: just before returning we can do a check that /bin/stunnel works correctly (w/ any shared libs that it needs).
If that fails, then we can warn and bail out

packaging/aznfs/DEBIAN/postinst
```diff
@@ -75,6 +131,14 @@ chmod 4755 /sbin/mount.aznfs
 mkdir -p /opt/microsoft/aznfs/data
 chmod 0755 /opt/microsoft/aznfs/data
 
+if command -v stunnel >/dev/null 2>&1; then
+    INSTALLED_STUNNEL_VERSION=$(stunnel -version 2>&1 | grep -Eo 'stunnel [0-9]+\.[0-9]+' | awk '{print $2}')
+fi
+
+if ! command -v stunnel >/dev/null 2>&1 || (( $(echo "$REQUIRED_STUNNEL_VERSION > $INSTALLED_STUNNEL_VERSION" | bc -l) )); then
```
linuxsmiths, line 138: see if we can avoid bc for comparing.

packaging/aznfs/DEBIAN/postinst
```diff
@@ -75,6 +131,14 @@ chmod 4755 /sbin/mount.aznfs
...
+    install_stunnel
+fi
```
linuxsmiths, line 140: else log a message saying stunnel version installed is same or newer than required and log both versions

packaging/aznfs/RPM/aznfs.spec
linuxsmiths, line 8: see if make can go to Required

scripts/aznfs_install.sh
```diff
@@ -280,7 +275,7 @@ ensure_pkg()
     local pkg="$1"
     local distro="$distro_id"
 
-    if [ "$distro" == "ubuntu" -o "$distro" == "debian" ]; then
```
linuxsmiths, line 283: revert this to be safe

src/mountscript.sh
```diff
@@ -37,6 +37,10 @@ get_host_from_share()
 {
     local hostshare="$1"
     local azprefix="$2"
+    local account=""
+    local hostprefix=""
+    local hostparts=()
```
linuxsmiths, line 42: declare -a hostparts

packaging/aznfs/DEBIAN/postinst
sroghanchi: replaced it with awk.

packaging/aznfs/DEBIAN/postinst
sroghanchi: wget should already be installed when it gets to postinst. gcc and make is also installed as requirements.

packaging/aznfs/RPM/aznfs.spec
sroghanchi: Moved make to requires. build-essential package doesn't exist on suse, so Recommends only installs the package if exists and not complain if it doesn't.

packaging/aznfs/DEBIAN/postinst
sroghanchi: I took another look on the stunnel installation. If there already is a stunnel installation, the binary is in **/bin/stunnel**. The stunnel from the package is installed in **/usr/local/bin/stunnel**. So after installing from source code, we should create the symlink for the stunnel to work fine.

I do need this line, otherwise the link doesn't work fine:

azureuser@sroghanchiubuntu22:~$ which stunnel
/usr/local/bin/stunnel
azureuser@sroghanchiubuntu22:~$ stunnel -version
-bash: /usr/bin/stunnel: No such file or directory

packaging/aznfs/DEBIAN/postinst
```diff
@@ -65,7 +61,20 @@ install_stunnel()
     fi
 
     cleanup_stunnel_files
-    sudo ln -sf /usr/local/bin/stunnel /bin/stunnel
+
+    # Remove the old link and create a symlink to stunnel binary.
+    [ -f /bin/stunnel ] && mv /bin/stunnel /bin/stunnel.old
+    rm -f /bin/stunnel
```
linuxsmiths, line 67: with the mv above this is a no-op

lib/common.sh
```diff
@@ -716,9 +716,6 @@ if command -v netstat &> /dev/null; then
     NETSTATCOMMAND="netstat"
 elif command -v ss &> /dev/null; then
     NETSTATCOMMAND="ss"
-else
-    eecho "[FATAL] Neither 'netstat' nor 'ss' is available!"
-    exit 1
 fi
```
linuxsmiths, line 719: Since you have removed this from here, you should be checking in v4 specific place.
I don't see that, or did u already have it?

linuxsmiths (approved):
pls look at the few comments I've added

lib/common.sh
sroghanchi: I already had it in both watchdogv4 and nfsv4mountscript. Ex in nfsv4mountscript:

  if [ -z "$NETSTATCOMMAND" ]; then
        eecho "No socket statistics command (netstat or ss) found! Cannot proceed with TLS mount."
        exit 1
    fi

packaging/aznfs/DEBIAN/postinst
sroghanchi: Removed it.

packaging/aznfs/RPM/aznfs.spec
```diff
@@ -90,6 +87,22 @@ if ! command -v stunnel >/dev/null 2>&1 || (( $(echo "$REQUIRED_STUNNEL_VERSION
 	fi
 
 	cleanup_stunnel_files
+
+	# Remove the old link and create a symlink to stunnel binary.
+    [ -f /bin/stunnel ] && mv /bin/stunnel /bin/stunnel.old
+    rm -f /bin/stunnel
```
linuxsmiths, line 93: forgot to address this?

linuxsmiths (approved):

packaging/aznfs/RPM/aznfs.spec
sroghanchi: Sorry, I forgot about this one. I'd only removed it from postint.

PR #150, opened 2024-11-18
 [BUG][release/aznfs2.0] Fixing BASH_ENV allows non-privileged users to execute commands with root privileges

palashvij-msft:


linuxsmiths (approved):

linuxsmiths:
why is this not meged yet?

src/mount.aznfs.c
```diff
@@ -13,6 +13,8 @@
 
 int main(int argc, char *argv[])
 {
+    unsetenv("BASH_ENV");
```
tlp-polyrize, line 16: One can still use the `LD_PRELOAD` environment variable.
Is using `sh` instead of `bash`, or using `execve` not an option?

PR #151, opened 2024-11-28
Update README.md

DeeptiJainMicrosoft:
Updated details as per my discussion with PG:

User must unmount all Azure Blob NFS share mount points before setting the AZNFS_IP_PREFIXES environment variable and wait for 5 minutes to clear the entry from mountmap file if any before remounting it.


linuxsmiths (approved):

linuxsmiths:
We should cherry pick this to release/aznfs2.0 too
@palashvij-msft

PR #152, opened 2024-12-11
Allow user to mount more than 20 accounts

palashvij-msft:


src/mountscript.sh
```diff
@@ -160,6 +163,31 @@ check_nconnect()
...
+                    fi
+
+                    # We reach here only if nconnect matches the regex 'matchstr="\<nconnect\>=([0-9]+)"'
+                    if [[ "$value" -lt 1 || "$value" -gt 16 ]]; then
+                        eecho "[ERROR] nconnect value must be provided between 1 and $AZNFS_MAX_NCONNECT"
+                        exit 1
+                    fi
+
+                    # Re-calculate the maximum accounts mountable from a single tenant.
+                    MAX_ACCOUNTS_MOUNTABLE_FROM_SINGLE_TENANT=$((320 / AZNFS_MAX_NCONNECT))
```
linuxsmiths, line 182: this much should move to L143, before the 2047/2048 check as the max nconnect value applies to both.

src/mountscript.sh
```diff
@@ -160,6 +163,31 @@ check_nconnect()
...
+
+                    # Check if the current nconnect value in use is suboptimal.
+                    if [[ "$value" -gt "$AZNFS_MAX_NCONNECT" ]]; then
```
linuxsmiths, line 185: this comment is for L210. There also we should make sure that we limit nconnect to min(4, $AZNFS_MAX_NCONNECT)

src/mountscript.sh
```diff
@@ -160,6 +163,31 @@ check_nconnect()
...
+                        pecho "Suboptimal nconnect value $value, forcing nconnect=$AZNFS_MAX_NCONNECT!"
```
linuxsmiths, line 186: this log should say limiting nconnect to such and such as advised by AZNFS_MAX_NCONNECT

src/mountscript.sh
linuxsmiths, line 182: you need to use this in find free address

linuxsmiths (commented):
this needs to be CP'ed to main too

src/mountscript.sh
palashvij-msft: this is the same variable that we use, directly updating here.

linuxsmiths (approved):

PR #157, opened 2025-01-10
Support ARM64 / AARCH deb-package

LUMERIIX:
I added a cross-compilation step to generate a ARM64 along the AMD64 deb-package using cross-compilation.

The current implementation may not be fully functional, but it allows you to build on AMD64 and ARM64 and get both artifacts. It should also be easily adaptable to add more platforms.

**Limitation**: You still get a FATAL_ERROR during [aznfs_install.sh](https://github.com/Azure/AZNFS-mount/blob/main/scripts/aznfs_install.sh) but the AZNFS setup still work wihtout updating capabilities.

### DIFF
--- a/package.sh
+++ b/package.sh
@@ -128,7 +128,7 @@ generate_tarball_package() {
 
 #STG_DIR, RELEASE_NUMBER and SOURCE_DIR will be taken as env var.
 pkg_name="aznfs"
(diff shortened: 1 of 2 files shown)


LUMERIIX:
@microsoft-github-policy-service agree company="Microsoft"

PR #168, opened 2025-03-12
Personal/sroghanchi/main pipeline

sroghanchi:
Package signing updates.

Example of yaml scripts for signing and publishing packages:
https://github.com/Azure/azure-storage-fuse/blob/main/blobfuse2-release.yaml#L1586-L1740

### DIFF
--- a/.github/aznfs-build.yaml
+++ b/.github/aznfs-build.yaml
@@ -33,337 +33,250 @@ pr:
   - none
 
 stages:
-  # - stage: BuildPackages
-  #   displayName: "Build and Package Artifacts"
-  #   jobs:
-  #     - job: Build_Package
-  #       displayName: "Build and Package"
-  #       pool:
-  #         vmImage: "ubuntu-latest"
-
-  #       variables:
-  #         - name: root_dir
-  #           value: '$(System.DefaultWorkingDirectory)'
-  #         - name: work_dir
-  #           value: '$(System.DefaultWorkingDirectory)/AZNFS-mount'
-
-  #       steps:
-  #         # - checkout: self
-
-  #         - script: |
-  #             git clone --branch personal/sroghanchi/pipeline https://github.com/Azure/AZNFS-mount
-  #           displayName: 'Checkout Code'
-  #           workingDirectory: $(root_dir)
-
-  #         # - script: |
-  #         #     echo "Building version ${{ parameters.versionName }}"
-  #         #   displayName: "Print Version Info"
-
-  #         - script: |
-  #             sudo apt-get update
-  #             sudo apt install -y gcc-aarch64-linux-gnu
-  #           displayName: "Install Cross-Compile Tools"
-
-  #         - script: |
-  #             export RELEASE_NUMBER=${{ parameters.versionName }}
-  #             export STG_DIR=$(work_dir)
-  #             export SOURCE_DIR=$(work_dir)
-  #             chmod +x $(work_dir)/generate_package.sh
-  #             $(work_dir)/generate_package.sh
-  #           displayName: "Run Package Script"
-
-  #         - script: |
-  #             mkdir -p $(Build.ArtifactStagingDirectory)/artifacts
-  #             mkdir -p $(Build.ArtifactStagingDirectory)/artifacts/deb
-  #             mkdir -p $(Build.ArtifactStagingDirectory)/artifacts/rpm
-  #             mkdir -p $(Build.ArtifactStagingDirectory)/artifacts/suse
-  #             mkdir -p $(Build.ArtifactStagingDirectory)/artifacts/stunnel
-  #             cp $(work_dir)/deb/aznfs-${{ parameters.versionName }}-1_amd64.deb $(Build.ArtifactStagingDirectory)/artifacts/deb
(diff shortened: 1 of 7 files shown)


.github/aznfs-build.yaml
```diff
@@ -0,0 +1,352 @@
...
+              cp $(work_dir)/suse/root/rpmbuild/RPMS/x86_64/aznfs-${{ parameters.versionName }}-1.x86_64.rpm $(Build.ArtifactStagingDirectory)/artifacts/suse
+              cp $(work_dir)/stunnel/root/rpmbuild/RPMS/x86_64/aznfs-${{ parameters.versionName }}-1.x86_64.rpm $(Build.ArtifactStagingDirectory)/artifacts/stunnel
+              echo "Listing Built Files..."
+              ls -R $(Build.ArtifactStagingDirectory)
+            displayName: "List Build Outputs"
+
+          - task: PublishBuildArtifacts@1
+            inputs:
+              PathtoPublish: $(Build.ArtifactStagingDirectory)
+              artifactName: 'aznfs-temp'
```
palashvij-msft, line 92: we'll be using aznfs later right? this is just for time being. Pls correct me if I am wrong

generate_package.sh
palashvij-msft, line 1: Let's discuss if we want to keep a seperate scipt or do it in the same.

.github/aznfs-build.yaml
palashvij-msft, line 1: Let's add this link as comment for reference for understanding terms-

https://learn.microsoft.com/en-us/azure/devops/pipelines/repos/azure-repos-git?view=azure-devops&tabs=yaml

.github/aznfs-build.yaml
```diff
@@ -0,0 +1,352 @@
...
+              sudo apt update 
+              sudo apt install apt-transport-https -y
+              sudo apt install dotnet-sdk-3.1 -y
+            displayName: "Update dependencies"
+
+          - task: EsrpCodeSigning@5
+            displayName: 'ESRP CodeSigning MI DEB'
+            inputs:
+              ConnectedServiceName: 'Azure AZNFS'
+              AppRegistrationClientId: '0eebccb1-4bcb-445e-bf3f-fa7a8fa5fd7d'
```
palashvij-msft, line 130: Is it okay to expose credentials like this? Better way would be to store using GitHub Secrets (best for Sensitive Data), also we would not be requiring to use it explicitly anywhere again in repo

.github/aznfs-build.yaml
palashvij-msft, line 1: It would be great if we could add few more comments in this yaml, will be easier for readers to understand in future as well.

palashvij-msft:
image

what is this?

lib/common.sh
```diff
@@ -626,7 +626,7 @@ log_version_info()
     elif [ "$distro_id" == "centos" -o "$distro_id" == "rocky" -o "$distro_id" == "rhel" ]; then
         current_version=$(yum info aznfs 2>/dev/null | grep "^Version" | tr -d " " | cut -d ':' -f2)
     elif [ "$distro_id" == "sles" ]; then
-        current_version=$(zypper info aznfs_sles 2>/dev/null | grep "^Version" | tr -d " " | cut -d ':' -f2 | cut -d '-' -f1)
+        current_version=$(zypper info aznfs 2>/dev/null | grep "^Version" | tr -d " " | cut -d ':' -f2 | cut -d '-' -f1)
```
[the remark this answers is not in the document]
sroghanchi: This way the aznfs package name is the same on all different distributions.

.github/aznfs-build.yaml
```diff
@@ -0,0 +1,352 @@
...
+          vmImage: "ubuntu-latest"
+
+        variables:
+          - name: root_dir
+            value: '$(System.DefaultWorkingDirectory)'
+          - name: work_dir
+            value: '$(System.DefaultWorkingDirectory)/AZNFS-mount'
+
+        steps:
+          # - checkout: self
```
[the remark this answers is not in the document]
sroghanchi: No, I'm removing it.

.github/aznfs-build.yaml
sroghanchi: No, that's a temporary directory that is created when pipeline runs.

sroghanchi:
> image
> what is this?

sroghanchi:
> image
> what is this?

This was a test pipeline that is disabled now.

.github/aznfs-build.yaml
```diff
@@ -0,0 +1,352 @@
...
+              inlineOperation: |
+                [
+                {
+                  "KeyCode": "CP-450779-Pgp",
+                  "OperationCode": "LinuxSign",
+                  "ToolName": "sign",
+                  "ToolVersion": "1.0",
+                  "Parameters": {}
+                }
+                ]
```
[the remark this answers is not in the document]
sroghanchi: The only thing I found so far is matrix, which I don't think it works for this case as for each matrix variable a new vm is created.

generate_package.sh
sroghanchi: We can discuss it. We can also keep both and after moving completely to PMC remove package.sh.

PR #171, opened 2025-03-18
Added build script

sroghanchi:


.github/aznfs-build.yaml
```diff
@@ -0,0 +1,352 @@
...
+          - script: |
+              sudo apt-get update
+              wget https://packages.microsoft.com/config/ubuntu/20.04/packages-microsoft-prod.deb
+              sudo dpkg -i packages-microsoft-prod.deb
+              sudo apt update 
+              sudo apt install apt-transport-https -y
+              sudo apt install dotnet-sdk-3.1 -y
+            displayName: "Update dependencies"
+
+          - task: EsrpCodeSigning@5
```
palashvij-msft, line 129: https://learn.microsoft.com/en-us/azure/devops/pipelines/process/templates?view=azure-devops&pivots=templates-includes

We can use the template here. We can call it signing-template.yml and provide parameters: folderPath, pattern, displayName to it.

call here in main yml using-

- template: signing-template.yml
  parameters:
    folderPath: $(Build.ArtifactStagingDirectory)/aznfs-temp/artifacts/deb
    pattern: '*.deb'
    displayName: 'ESRP CodeSigning MI DEB'
    
I don't see another way of doing it, maybe looping but that wouldn't be so clean too.

sroghanchi: I'm afraid that adding another yaml file only for the signing part might make it more complicated to follow or update.

linuxsmiths (approved):

packages.csv
```diff
@@ -0,0 +1,14 @@
+# Do not remove these comments - the list should start at line #3
+# Format of the file : <Distro Name>,<Architecture, File Type>,<Repo Name>,<Release Name(for deb repos only)>
```
linuxsmiths, line 2: 2nd column is arch or file type?
How do we specify arm64 f.e.?

.github/aznfs-build.yaml
```diff
@@ -0,0 +1,352 @@
...
+                  aznfsids=`pmc --base-url "https://pmc-ingest.trafficmanager.net/api/v4" package rpm list --name aznfs`
+                  echo "AZNFS rpm list ==========: $aznfsids"
+
+                  # upload packages
+                  pmc --base-url "https://pmc-ingest.trafficmanager.net/api/v4" package upload $(Build.ArtifactStagingDirectory)/aznfs-signed/artifacts
+
+                  #################################### get package ids #####################################
+
+                  aznfsDeb=`pmc --base-url "https://pmc-ingest.trafficmanager.net/api/v4" --id-only package upload $(Build.ArtifactStagingDirectory)/aznfs-signed/artifacts/deb/aznfs*-1_amd64.deb`
+                  echo "AZNFS DEB Package ID: $aznfsDeb"
```
linuxsmiths, line 326: we will need to support arm64 also

PR #189, opened 2025-03-27
Annotated main sample config too with range for all values.

linuxsmiths:


nitin-deamon (approved):

turbonfs/sample-turbo-config.yaml
```diff
@@ -12,42 +16,75 @@ container: nfsv3test
...
+#
+#filecache.enable: false
+#filecache.cachedir: /mnt
+#filecache.max_size_gb: 1000
+#cache_max_mb: 4096
+
+#
+# Config controlling misc system behaviour.
+#
+sys.force_stable_writes: true
```
nitin-deamon, line 40: On first thought shall we move another section Phase 1.5 
But on second thought for one config, it doesn't worth to create new section.

PR #192, opened 2025-03-28
Updated build script

sroghanchi:


packages.csv
```diff
@@ -4,6 +4,7 @@ Ubuntu-18.04,aznfsDeb,microsoft-ubuntu-bionic-prod-apt,bionic
 Ubuntu-20.04,aznfsDeb,microsoft-ubuntu-focal-prod-apt,focal
 Ubuntu-22.04,aznfsDeb,microsoft-ubuntu-jammy-prod-apt,jammy
 Ubuntu-24.04,aznfsDeb,microsoft-ubuntu-noble-prod-apt,noble
+RHEL-7.3,aznfsRpmStunnel,microsoft-rhel7.3-prod-yum,
 RHEL-7.0,aznfsRpmStunnel,microsoft-rhel7.0-prod-yum,
```
linuxsmiths, line 8: is this what we want?
I thought we need to use the 7.3 repo for 7.0 RHEL as well

linuxsmiths: this is more of a correctness thing, but I think what you have will work, as long as it uploads to rhel/7, that's what is the repo added

linuxsmiths (approved):

PR #193, opened 2025-03-28
Personal/sroghanchi/test fix certificate

sroghanchi:
- Fix certificate pinning
- Add TIMEOUTidle to stunnle config. This value has been tested for the past month in preprod.

### DIFF
--- a/src/nfsv4mountscript.sh
+++ b/src/nfsv4mountscript.sh
@@ -112,32 +112,89 @@ get_cert_path_based_and_command()
     if command -v apt-get &> /dev/null; then
         CERT_PATH="/usr/local/share/ca-certificates"
         CERT_UPDATE_COMMAND="update-ca-certificates"
-        STUNNEL_CAFILE="/etc/ssl/certs/DigiCert_Global_Root_G2.pem"
     # Check if we're on a Red Hat-based distribution
     elif command -v yum &> /dev/null || command -v dnf &> /dev/null; then
         CERT_PATH="/etc/pki/ca-trust/source/anchors"
         CERT_UPDATE_COMMAND="update-ca-trust extract"
-        STUNNEL_CAFILE="${CERT_PATH}/DigiCert_Global_Root_G2.crt"
+        mkdir -p /etc/ssl/certs
+        if [ $? -ne 0 ]; then
+            eecho "[FATAL] Not able to create /etc/ssl/certs path for certificate!"
+            return 1
+        fi
     # Check if we're on a SUSE-based distribution
     elif command -v zypper &> /dev/null; then
         CERT_PATH="/etc/pki/trust/anchors"
         CERT_UPDATE_COMMAND="update-ca-certificates"
-        STUNNEL_CAFILE="${CERT_PATH}/DigiCert_Global_Root_G2.crt"
+        mkdir -p /etc/ssl/certs
+        if [ $? -ne 0 ]; then
+            eecho "[FATAL] Not able to create /etc/ssl/certs path for certificate!"
+            return 1
+        fi
     else
         eecho "[FATAL] Unsupported distribution!"
         return 1
     fi
+
+    STUNNEL_CAFILE="/etc/ssl/certs/DigiCert_Global_Root_G2.pem"
+}
+
+extract_CA()
+{
+    awk '/DigiCert Global Root G2/ {found=1} found && /BEGIN CERTIFICATE/,/END CERTIFICATE/ {print > "'$STUNNEL_CAFILE'"} found && /END CERTIFICATE/ {exit}' /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem
+    if [ $? -ne 0 ]; then
+        eecho "[FATAL] Failed to extract DigiCert Global Root G2 certificate from /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem!"
+        return 1
+    fi
+}
+
+compare_CA_thumbprint()
+{
+    local thumbprint=$(openssl x509 -in $STUNNEL_CAFILE -noout -fingerprint | cut -d'=' -f2)
+    local expected_thumbprint=$(awk '/DigiCert Global Root G2/ {found=1} found && /BEGIN CERTIFICATE/,/END CERTIFICATE/ {print} found && /END CERTIFICATE/ {exit}' /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem | openssl x509 -noout -fingerprint -sha1 2&>1 | cut -d'=' -f2)
+
+    if [ "$thumbprint" != "$expected_thumbprint" ]; then
+        return 1
+    fi
+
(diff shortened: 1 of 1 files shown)


src/nfsv4mountscript.sh
```diff
@@ -112,32 +114,89 @@ get_cert_path_based_and_command()
...
+        if [ $? -ne 0 ]; then
+            eecho "[FATAL] Not able to create /etc/ssl/certs path for certificate!"
+            return 1
+        fi
     # Check if we're on a SUSE-based distribution
     elif command -v zypper &> /dev/null; then
         CERT_PATH="/etc/pki/trust/anchors"
         CERT_UPDATE_COMMAND="update-ca-certificates"
-        STUNNEL_CAFILE="${CERT_PATH}/DigiCert_Global_Root_G2.crt"
+        mkdir -p /etc/ssl/certs
```
zimchen-msft, line 130: below you have a comment mentioning the suse use this folder as default root ca folder, that case why we need to try create it here? isn't it be similar like in ubuntu that you don't need to do this?

src/nfsv4mountscript.sh
```diff
@@ -112,32 +114,89 @@ get_cert_path_based_and_command()
...
+compare_CA_thumbprint()
+{
+    local thumbprint=$(openssl x509 -in $STUNNEL_CAFILE -noout -fingerprint 2>/dev/null | cut -d'=' -f2)
+    local expected_thumbprint=$(awk '/DigiCert Global Root G2/ {found=1} found && /BEGIN CERTIFICATE/,/END CERTIFICATE/ {print} found && /END CERTIFICATE/ {exit}' /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem | openssl x509 -noout -fingerprint -sha1 2>/dev/null | cut -d'=' -f2)
+
+    if [ "$thumbprint" != "$expected_thumbprint" ]; then
+        return 1
+    fi
+
+    vecho "Thumbprint of the installed DigiCert Global Root G2 certificate matches the expected value."
```
zimchen-msft, line 161: log both thumbprints if not match.

src/nfsv4mountscript.sh
```diff
@@ -153,13 +212,28 @@ add_stunnel_configuration()
     fi
 
     if [ ! -f $STUNNEL_CAFILE ]; then
-        vecho "CA root cert is missing for stunnel configuration. Installing DigiCert_Global_Root_G2 certificate."
+        vecho "CA root cert is missing for stunnel configuration. Install or extract DigiCert_Global_Root_G2 certificate."
         install_CA_cert
```
zimchen-msft, line 216: the logic around and inside `install_CA_cert` is expecting single thread. I think we should synchronize start from line 214, because if user have multiple account configured in fstab, then the `mount -a` can cause some race condition that may contaminate the `DigiCert_Global_Root_G2.pem`

zimchen-msft (changes requested):

src/nfsv4mountscript.sh
sroghanchi: Right, I removed it.

src/nfsv4mountscript.sh
sroghanchi: We are already holding the mountmapv4 file lock while creating the stunnel config file, so two mount script processes won't run this part of code at the same time.

src/nfsv4mountscript.sh
```diff
@@ -112,32 +114,88 @@ get_cert_path_based_and_command()
...
+            eecho "[FATAL] Not able to create /etc/ssl/certs path for certificate!"
+            return 1
+        fi
     # Check if we're on a SUSE-based distribution
     elif command -v zypper &> /dev/null; then
         CERT_PATH="/etc/pki/trust/anchors"
         CERT_UPDATE_COMMAND="update-ca-certificates"
-        STUNNEL_CAFILE="${CERT_PATH}/DigiCert_Global_Root_G2.crt"
+        if [ $? -ne 0 ]; then
+            eecho "[FATAL] Not able to create /etc/ssl/certs path for certificate!"
```
zimchen-msft, line 131: is this `if [ $? -ne 0 ]; then` still need?

src/nfsv4mountscript.sh
```diff
@@ -112,32 +114,88 @@ get_cert_path_based_and_command()
...
+    fi
+}
+
+compare_CA_thumbprint()
+{
+    local thumbprint=$(openssl x509 -in $STUNNEL_CAFILE -noout -fingerprint 2>/dev/null | cut -d'=' -f2)
+    local expected_thumbprint=$(awk '/DigiCert Global Root G2/ {found=1} found && /BEGIN CERTIFICATE/,/END CERTIFICATE/ {print} found && /END CERTIFICATE/ {exit}' /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem | openssl x509 -noout -fingerprint -sha1 2>/dev/null | cut -d'=' -f2)
+
+    if [ "$thumbprint" != "$expected_thumbprint" ]; then
+        return 1
```
zimchen-msft, line 157: still not log the hash if mismatch.
move line 160 to 155

src/nfsv4mountscript.sh
sroghanchi: Done.

src/nfsv4mountscript.sh
```diff
@@ -112,32 +114,88 @@ get_cert_path_based_and_command()
...
+    fi
+
+    vecho "Thumbprint of the installed DigiCert Global Root G2 certificate matches the expected value. Expected: ${expected_thumbprint}, Installed: ${thumbprint}."
 }
 
 install_CA_cert()
 {
-    wget https://cacerts.digicert.com/DigiCertGlobalRootG2.crt.pem --no-check-certificate -O ${CERT_PATH}/DigiCert_Global_Root_G2.crt
+    # For Debian-based and SUSE-based distributions, if the cert exits, it's in /etc/ssl/certs/DigiCert_Global_Root_G2.pem. For RedHat-based distributions, it's' in /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem.
+    if command -v yum &> /dev/null || command -v dnf &> /dev/null; then
```
zimchen-msft, line 166: looks like we use the distro check more than previous. I think it would be good to do the distro check at the beginning of `tls_nfsv4_files_share_mount` and then saved the value to global `isRedHat`, `isDebian`, `isSUSE` and use it everywhere else
up to you

zimchen-msft (approved):

src/nfsv4mountscript.sh
sroghanchi: Done.

src/nfsv4mountscript.sh
sroghanchi: Done.

src/nfsv4mountscript.sh
sroghanchi: Done.

linuxsmiths (approved):
Approving w/o review.

PR #194, opened 2025-03-28
Fix build script to properly read packages.csv

sroghanchi:


.github/aznfs-build.yaml
```diff
@@ -344,14 +344,14 @@ stages:
                   ################################### Add uploaded packages to its repository ######################################
                   echo "now adding packages to repositories..."
 
-                  while IFS=, read -r distro PackageArchType repoName releaseName; do
+                  while IFS=, read -r distro PackageArchType repoName releaseName || [ -n "$distro" ]; do
                     echo "Uploading packages for $distro"
                     pmc --base-url "https://pmc-ingest.trafficmanager.net/api/v4" repo package update --add-packages ${!PackageArchType} $repoName $releaseName
                   done < <(tail -n +3 $(Build.ArtifactStagingDirectory)/packages.csv)
```
linuxsmiths, line 351: in future if packages.csv is saved with a newline (if git saves it with a newline) this will try to run pmc with empty parameters, so following would be more robust.

```
while IFS=, read -r distro PackageArchType repoName releaseName && [ -n "$repoName" ]; do
...
done < <(tail -n +3 $(Build.ArtifactStagingDirectory)/packages.csv; echo)
```

it forces a newline and then only enters the while loop for valid lines

sroghanchi: Addressed, thanks!

linuxsmiths (approved):

PR #223, opened 2025-04-29
Relax FQDN validation for NFSv4

sroghanchi:


src/mountscript.sh
```diff
@@ -57,23 +57,25 @@ get_host_from_share()
         return 1
     fi
 
-    # Split host by "."
-    IFS=. read -r -a hostparts <<< "$host"
+    if [ "$nfs_vers" == "3" ]; then
```
zimchen-msft, line 60: what about `vers == 4.1 && -n $AZURE_ENDPOINT_OVERRIDE`?
if that condition is not needed, then do we still need keep the changes in line 67 for xnfs?

sroghanchi: I think we can skip this check for v4 if we are relaxing the whole FQDN validation. This check happens before is_valid_fqdn and only checks the version and if the FQDN is a blob endpoint or file endpoint. If the user has forgotten to set AZURE_ENDPOINT_OVERRIDE, I think it's better to emit the error where we check FQDN. In this function error is only complaining about the version incompatibility with "file" or "blob" in the endpoint.

I removed "file" from line 67.

README.md
```diff
@@ -98,9 +98,9 @@ AZNFS is supported on following Linux distros:
 	```
 	sudo mount -t aznfs -o vers=4.1 <account-name>.file.core.windows.net:/<account-name>/<container-name> /mountpoint
 	```
-   For isolated environments, ensure that the environment variable "AZURE_ENDPOINT_OVERRIDE" is set to the appropriate endpoint before running the mount command:
+   If using isolated environments, or using custom DNS in any cloud other that the public cloud, ensure that the environment variable "AZURE_ENDPOINT_OVERRIDE" is set to the appropriate endpoint before running the mount command. For example:
```
RenaShahMSFT, line 101: Remember to set environment variable "AZURE_ENDPOINT_OVERRIDE" for mounting non-Public Azure Cloud regions and when using Custom DNS. For example: export AZURE_ENDPOINT_OVERRIDE="chinacloudapi.cn" for Azure China Cloud.

src/mountscript.sh
```diff
@@ -245,14 +247,19 @@ if [ "$USING_AZNFSCLIENT" != true ] || [ "$1" != "none" ]; then
 
     # TODO: Comment out below code for devfabric. 'is_valid_fqdn' will fail on devfabric.
     if ! is_valid_fqdn "$nfs_host" "$AZ_PREFIX"; then
-        eecho "Not a valid Azure $AZ_PREFIX NFS endpoint: ${nfs_host}!"
-        if [[ -n "$AZURE_ENDPOINT_OVERRIDE" ]]; then
-            eecho "Must be of the form 'account.$AZ_PREFIX.core.$AZURE_ENDPOINT_OVERRIDE'!"
+        if [ "$nfs_vers" == "4.1" ]; then
+            wecho "Not a valid Azure $AZ_PREFIX NFS endpoint: ${nfs_host}!"
+            wecho "If using an isolated environemnt, or using a cusotm DNS and your account is not hosted on public cloud, please set the environment variable AZURE_ENDPOINT_OVERRIDE to the appropriate endpoint suffix!"
         else
```
RenaShahMSFT, line 253: " $AZ_PREFIX NFS endpoint: ${nfs_host} is a not default storage-account FQDN. Additional step may be needed to mount"
            wecho "Remember to set environment variable "AZURE_ENDPOINT_OVERRIDE" for mounting non-Public Azure Cloud regions or when using Custom DNS."

RenaShahMSFT (commented):
Left some suggested messages

zimchen-msft (approved):

src/mountscript.sh
```diff
@@ -57,23 +57,25 @@ get_host_from_share()
...
 
-    # Check if the prefix matches the expected azprefix
-    if [ "$hostprefix" != "$azprefix" ]; then
-        echo "Bad share name: ${hostshare}."
-        echo "Share must be of the form 'account.$azprefix.core.windows.net:/account/container' for vers=$nfs_vers"
-        return 1
+        # Check if the prefix matches the expected azprefix
+        if [ "$hostprefix" != "$azprefix" ]; then
+            echo "Bad share name: ${hostshare}."
+            echo "Share must be of the form 'account.$azprefix.core.windows.net:/account/container' for vers=$nfs_vers"
```
linuxsmiths, line 76: Since we suppot oher forms of fqdn, I think we should emit a more generic error here. Also, the log at L56 should be same as this.
Also the "for vers=" is not very helpful.

```
echo "Share to be mounted must be of the form '<endpoint fqdn>:/account/container', where most common endpoint fqdn is 'account.$azprefix.core.windows.net', but can have one of the other supported forms"
```

src/mountscript.sh
```diff
@@ -57,23 +57,25 @@ get_host_from_share()
...
-       if [[ "$part" == "file" || "$part" == "blob" ]]; then
-            hostprefix="$part"
-            break
-        fi
-    done
+        for part in "${hostparts[@]}"; do
+        if [[ "$part" == "blob" ]]; then
+                hostprefix="$part"
+                break
+            fi
```
linuxsmiths, line 70: indent 4 spaces to the right

linuxsmiths (approved):
RLG

PR #227, opened 2025-05-06
[release/3.0] CP-update for oracle linux

palashvij-msft:


lib/common.sh
```diff
@@ -623,7 +623,7 @@ log_version_info()
 
     if [ "$distro_id" == "ubuntu" ]; then
         current_version=$(dpkg-query -W -f='${Version}\n' aznfs 2>/dev/null)
-    elif [ "$distro_id" == "centos" -o "$distro_id" == "rocky" -o "$distro_id" == "rhel" ]; then
+    elif [ "$distro_id" == "centos" -o "$distro_id" == "rocky" -o "$distro_id" == "rhel"  -o "$distro" == "mariner" -o "$distro" == "ol" ]; then
```
linuxsmiths, line 626: is this sufficient for mariner support?
also, I don't remember this was part of 2.0 change?

scripts/aznfs_install.sh
```diff
@@ -134,7 +134,7 @@ ensure_pkg()
 {
     local distro="$distro_id"
 
-    if [ "$distro" == "ubuntu" -o "$distro" == "debian" ]; then
+    if [ "$distro" == "ubuntu" ]; then
```
linuxsmiths, line 137: who had added this and why did you remove this?

linuxsmiths (approved):
RLG

lib/common.sh
palashvij-msft: I have not yet validated Mariner, as I noticed this at the last moment. After reviewing packages.csv, it appears we are not uploading the packages to Mariner yet. I added this to ensure consistency once, need to update the PRs one more iteration.

scripts/aznfs_install.sh
palashvij-msft: I didn't add this, and iirc, I asked sepideh in past to remove this too if we are not using. I have put this PR up for consistency with release/aznfs2.0 branch once, will check with her and take action as needed. not merging for now.

PR #233, opened 2025-05-26
Allow 'fuse' option too along with turbo option

shubham808:


src/mountscript.sh
```diff
@@ -186,6 +186,14 @@ check_turbo_option()
     if [[ "$MOUNT_OPTIONS" =~ $matchstr ]]; then
         export USING_AZNFSCLIENT=true
     fi
+    
+    #
+    # Check if fuse flag is passed. We need to support both options for some reason.
```
nitin-deamon, line 191: We need to support both options as some customer using turbo option.
TODO: Once we have confirmation no customer using should remove turbo option.

nitin-deamon (approved):

PR #237, opened 2025-06-23
Personal/sroghanchi/notlsversion

sroghanchi:
Remove default TLS1.2 option and add a mount option for TLS version.

### DIFF
--- a/src/nfsv4mountscript.sh
+++ b/src/nfsv4mountscript.sh
@@ -251,14 +251,6 @@ add_stunnel_configuration()
         return 1
     fi
 
-    # TODO: Change to TLSv1.3 once we have TLSv1.3 version enabled.
-    echo "sslVersion = TLSv1.2" >> $stunnel_conf_file
-    if [ $? -ne 0 ]; then
-        chattr -f +i $stunnel_conf_file
-        eecho "Failed to add sslVersion option to $stunnel_conf_file!"
-        return 1
-    fi
-
     echo "debug = $DEBUG_LEVEL" >> $stunnel_conf_file
     if [ $? -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


src/nfsv4mountscript.sh
```diff
@@ -396,6 +395,27 @@ tls_nfsv4_files_share_mount()
...
+
+        # Check if SSL version is either 1.2 or 1.3
+        if [[ "$ssl_version" == "1.2" || "$ssl_version" == "1.3" ]]; then
+            echo "TLS version option: $ssl_version"
+        else
+            echo "No valid TLS version. Please provide a valid TLS version (1.2 or 1.3)."
+            exit 1
+        fi
+
+        # Remove the tls option from MOUNT_OPTIONS.
```
zimchen-msft, line 411: The remove of the new option should happen for notls too, basically, if user accidentally use notls and tls=1.3 together, we better not show errors, because tls version is not something we would like to mention in GA doc, so if both exist, notls should hold the priority.

zimchen-msft: Basically, right before line 853 execute, check and remove all new options we added.
Also add logs for these corner cases for icm debug purpose.

zimchen-msft (changes requested):

zimchen-msft (approved):

src/nfsv4mountscript.sh
sroghanchi: If a user specifies both notls and tls=x, they correctly receive an "invalid mount option" error. I believe it’s best to explicitly disallow using these options together, since the combination is invalid and could mislead users into thinking they’re mounting with tls=x when notls is also present.

PR #240, opened 2025-06-24
Only run pipeline from non-main branches

sroghanchi:


.github/aznfs-build.yaml
```diff
@@ -29,6 +29,7 @@ variables:
 stages:
   - stage: BuildPackages
     displayName: "Build and Package Artifacts"
+    condition: ne(variables['Build.SourceBranch'], 'refs/heads/main')
```
palashvij-msft, line 32: It restricts running from main, but what if someone wants to test changes from main itself?
we should add some conditions like-
1> if main/non-main, then make sure release number should be less than 3.0.0
2> allow release/* branch for all versions

Let me know your thoughts?

palashvij-msft: We did something similar check in release.yaml for github pipeline too, can refer.

sroghanchi: I see. I was thinking that test changes can be run from the test branches and not main, but you are right we should also restrict the release version from non-release branches.

.github/aznfs-build.yaml
```diff
@@ -27,7 +27,31 @@ variables:
 # ADO pipelines documentation:
 # https://learn.microsoft.com/en-us/azure/devops/pipelines/repos/azure-repos-git?view=azure-devops&tabs=yaml
 stages:
+  - stage: CheckVersion
```
palashvij-msft, line 30: https://learn.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands
   
    - job: CheckVersion
      steps:
      - script: |
          # Enable strict error handling for safer script execution
          set -euo pipefail

          echo "##[section]Version validation script started"
          echo "Branch ref: $(Build.SourceBranch)"
          echo "Provided version: ${{ parameters.versionName }}"

          # Extract the actual branch name from the full reference (e.g., refs/heads/feature/foo -> feature/foo)
          branchName="$(echo "$(Build.SourceBranch)" | sed 's#refs/heads/##')"
          echo "Extracted branch name: $branchName"

          # Define the release branch and minimum allowed version for flexibility
          release_branch="release/3.0"
          min_version="3.0.0"

          # Only validate version if NOT on the release branch
          if [[ "$branchName" != "$release_branch" ]]; then
            version="${{ parameters.versionName }}"
            echo "##[command]Non-release branch detected. Checking if version $version exceeds allowed minimum ($min_version)..."

            # Compare version using natural version sort
            if [[ "$(echo -e "$version\n$min_version" | sort -V | head -n1)" != "$version" ]]; then
              echo "##[error]Version $version is NOT permitted on non-release branches (must be <= $min_version)."
              exit 1
            else
              echo "##[command]Version check PASSED: $version is allowed on $branchName."
            fi
          else
            echo "##[command]On release branch ($release_branch). Version check skipped."
          fi

          echo "##[section]Version validation script completed"
        displayName: 'Validate versionName for non-release branches'

PR #245, opened 2025-07-01
[BUG][release/3.0] [new] update issue with package dependency eval.

palashvij-msft:


packaging/aznfs/RPM/aznfs.spec
```diff
@@ -11,6 +11,9 @@ Recommends: build-essential
 Requires: bash, PROCPS_PACKAGE_NAME, conntrack-tools, iptables, bind-utils, iproute, util-linux, nfs-utils, NETCAT_PACKAGE_NAME, newt, stunnel, net-tools
 %endif
 
+%global __provides_exclude_from ^/opt/microsoft/aznfs/libs/.*\.so.*$
+%global __requires_exclude_from ^(/opt/microsoft/aznfs/libs/.*\.so.*|/sbin/aznfsclient)$
```
linuxsmiths, line 15: Add a comment explaining why this is needed, else new users reading it after some time won't understand and they cannot extend/maintain it properly.

```
#
# We bundle some libs under /opt/microsoft/aznfs/libs/ which are to be used by /sbin/aznfsclient.
# This allows us to not have dependency on the libs provided by the distro thus making us agnostic of the distro and the
# same binary correctly works on all distros regardless of the glibc version (and/or other libs) used by the distro.
# We do this as some of the libs that we use do not have a static version.
# With the __provides_exclude_from directive we tell the RPM package manager to leave these libs alone for our use
# and not confuse other packages with those. Since we are self sufficient and we don't need any libs from the system,
# we use the __requires_exclude_from directive to tell the package manager.
#
```

PR #246, opened 2025-07-02
Update README.md

taoyama:
Those should be RHEL instead of RedHat.
There were Red Hat Linux distributions before Red Hat *Enterprise* Linux.


taoyama:
Reference: https://en.wikipedia.org/wiki/Red_Hat_Linux

PR #249, opened 2025-07-16
[NFS] Added authnone changes

pragyagandhi:


turbonfs/src/main.cpp
```diff
@@ -501,6 +501,27 @@ int get_authinfo_data(struct auth_info& auth_info)
...
+
+    // Allocate response structure
+    auth_token_cb_res *cb_res = (auth_token_cb_res *) malloc(sizeof(auth_token_cb_res));    
+    if (!cb_res) {
+        AZLogError("Failed to allocate memory for auth_token_cb_res");
+        return nullptr;
+    }
+
+    cb_res->azauth_data = strdup("None");
+    cb_res->expiry_time = static_cast<uint64_t>(time(NULL))+300;
```
nitin-deamon, line 520: why 300 we should remove it

turbonfs/inc/nfs_internal.h
```diff
@@ -108,7 +108,8 @@ struct mount_options
         actimeo(aznfsc_cfg.actimeo),
         readdir_maxcount(aznfsc_cfg.readdir_maxcount),
         readahead_kb(aznfsc_cfg.readahead_kb),
-        auth(aznfsc_cfg.auth)
+        auth(aznfsc_cfg.auth),
+        authtype(aznfsc_cfg.authtype)
```
nitin-deamon, line 112: Why we need to set authType as aznfsc_cfg has default value.
We can set in constructor according to auth
authtype = AzAuthNone;
If (auth)
 authtype = AzAuthAAD;

turbonfs/src/connection.cpp
```diff
@@ -60,19 +59,27 @@ std::string get_clientid() {
...
+    static std::string fallback_client_id = std::to_string(get_current_usecs()) + "-unknown";
+    AZLogDebug("Using fallback clientid {}", fallback_client_id);
+    return fallback_client_id;
 }
 
 bool nfs_connection::open()
 {
     const int nodelay = 1;
+    std::string client_id;
+    int ret = -1;
```
nitin-deamon, line 81: revert it back to original declare it at time of use.

turbonfs/src/connection.cpp
```diff
@@ -100,38 +107,39 @@ bool nfs_connection::open()
...
-            AZLogError("Failed to set auth values in nfs context, "
-                       "exportpath={} authtype={} "
-                       "clientversion={} clientid={}",
-                       mo.export_path.c_str(),
-                       mo.authtype.c_str(),
-                       client_version,
-                       client_id.c_str());
-            goto destroy_context;
-        }
+        mo.authtype = "AzAuthAAD";
```
nitin-deamon, line 110: Its set in constructor then we don;t need it.

turbonfs/src/connection.cpp
```diff
@@ -100,38 +107,39 @@ bool nfs_connection::open()
...
+    // 16 should be sufficient to hold the version string.
+    char client_version[16];
+
+    n = snprintf(client_version, sizeof(client_version),
+                                "%d.%d.%d", AZNFSCLIENT_VERSION_MAJOR,
+                                AZNFSCLIENT_VERSION_MINOR,
+                                AZNFSCLIENT_VERSION_PATCH);
+    assert(n < sizeof(client_version));
+
+    client_id = get_clientid();
```
nitin-deamon, line 122: static const std::string client_id = get_clientid();

PR #277, opened 2025-10-29
RA-GRS support for turboNFS

palashvij-msft:
1>
For an RAGRS account, looks like we are mounting the primary endpoint
 
root@amnnfswestus8:~# mount -t aznfs -o vers=3,proto=tcp,fuse nuaxscnbn9pe18cx-secondary.blob.preprod.core.windows.net:/nuaxscnbn9pe18cx/cont1 /mnt/nuaxscnbn9pe18cx_ragrs_ra
root@amnnfswestus8:/mnt/nuaxscnbn9pe18cx_ragrs_ra# touch abc
root@amnnfswestus8:/mnt/nuaxscnbn9pe18cx_ragrs_ra# ls
abc  amn_write  amn_writeq

 **ideal behaviour as shown with native nfs client**
root@amnnfswestus8:~# mount -t  nfs -o vers=3,proto=tcp nuaxscnbn9pe18cx-secondary.blob.preprod.core.windows.net:/nuaxscnbn9pe18cx/cont1 /mnt/nuaxscnbn9pe18cx_ragrs_ra

root@amnnfswestus8:/mnt/nuaxscnbn9pe18cx_ragrs_ra# touch agd
touch: cannot touch 'agd': Read-only file system

2> It is letting you mount even a non-ragrs account without throwing any error but it mounts the primary endpoint

### DIFF
--- a/turbonfs/sample-turbo-config.yaml
+++ b/turbonfs/sample-turbo-config.yaml
@@ -8,6 +8,9 @@
 # "none" as source in the mount command. In that case account and container
 # are mandatory, cloud_suffix can be guessed and port is default 2048.
 #
+# For accessing the secondary endpoint (RA-GRS), append "-secondary" to the
+# account name, e.g., "myaccount-secondary".
+#
 account: sjc22prdste06hnfsv3acc1
 container: nfsv3test
 


turbonfs/src/config.cpp
```diff
@@ -445,7 +445,14 @@ bool aznfsc_cfg::set_defaults_and_sanitize()
 
     // Set aggregates.
     server = std::string(account) + "." + std::string(cloud_suffix);
-    export_path = "/" + std::string(account) + "/" + std::string(container);
+    if (std::string(account).size() > 10 &&
```
nitin-deamon: if (int n = std::string(account).find("-secondary"); n != std::string(account).size()) {
	export_path = "/" + account.substr(0,n) + "/" + std::string(container);
}

pragyagandhi (approved):

turbonfs/src/config.cpp
```diff
@@ -445,7 +445,11 @@ bool aznfsc_cfg::set_defaults_and_sanitize()
 
     // Set aggregates.
     server = std::string(account) + "." + std::string(cloud_suffix);
-    export_path = "/" + std::string(account) + "/" + std::string(container);
+    if (int n = std::string(account).find("-secondary"); n != std::string::npos) {
```
amulyan13: Can you make changes to the sample-turbo-config.yaml as well to add a comment for how account name should be set for secondary account?


//# Share details
//# These will ONLY be used if the user runs the binary directly or passes
// "none" as source in the mount command. In that case account and container
//are mandatory, cloud_suffix can be guessed and port is default 2048.
//
account: sjc22prdste06hnfsv3acc1

PR #292, opened 2026-02-02
Add nfsfs servers guard to prevent premature cleanup after unmount

bharathsm-ms:
Add a guard that checks /proc/fs/nfsfs/servers ip:port before cleaning up stunnel to avoid premature teardown.
In AKS, we observed a case where the pod’s mount namespace was still referenced by another process even after the host unmounted it. When the host unmount happened, stunnel was cleaned up and the NFS connection got stuck in a bad state, which could result in hung mounts. This fix handle such cases by actually checking for servers reference before cleaning up stunnel after unmount operation.

### DIFF
--- a/src/aznfswatchdogv4
+++ b/src/aznfswatchdogv4
@@ -21,6 +21,12 @@ ip_to_hex()
     if [ -z "$o1" -o -z "$o2" -o -z "$o3" -o -z "$o4" ]; then
         return 1
     fi
+    if ! [[ "$o1" =~ ^[0-9]+$ && "$o2" =~ ^[0-9]+$ && "$o3" =~ ^[0-9]+$ && "$o4" =~ ^[0-9]+$ ]]; then
+        return 1
+    fi
+    if [ "$o1" -gt 255 -o "$o2" -gt 255 -o "$o3" -gt 255 -o "$o4" -gt 255 ]; then
+        return 1
+    fi
     printf "%02x%02x%02x%02x" "$o1" "$o2" "$o3" "$o4"
 }
 
@@ -33,6 +39,14 @@ is_nfs_server_active_for_target()
 
     if [ -z "$target_ip" -o -z "$target_port" ]; then
         vecho "NFSv4 server entry check: missing target ip/port (ip='$target_ip', port='$target_port')."
+        return 1
+    fi
+    if ! [[ "$target_port" =~ ^[0-9]+$ ]]; then
+        vecho "NFSv4 server entry check: invalid target port (port='$target_port')."
+        return 1
+    fi
+    if [ "$target_port" -lt 1 -o "$target_port" -gt 65535 ]; then
+        vecho "NFSv4 server entry check: target port out of range (port='$target_port')."
         return 1
     fi
 


src/aznfswatchdogv4
```diff
@@ -10,6 +10,82 @@
...
+        local version_field=${fields[0]}
+        local server_field=${fields[1]}
+        local port_field=${fields[2]}
+
+        # Only match NFSv4 entries.
+        if [ "$version_field" != "v4" ]; then
+            continue
+        fi
+
+        # /proc/fs/nfsfs/servers uses hex format for server and port.
```
zimchen-msft: Does the server and port are always using hex for different kernel versions? for example, is it possible some old kernel versions are using decimal and newer one using hex?

bharathsm-ms: Almost since the introduction of /proc/fs/nfsfs/servers file (18+ years) these are displayed in hex format.
(https://elixir.bootlin.com/linux/v6.19-rc5/source/fs/nfs/client.c#L1375)

zimchen-msft (approved):

msftdan (approved):
LGTM

src/aznfswatchdogv4
```diff
@@ -10,6 +10,82 @@
...
+
+    if [ -z "$target_ip_hex" ]; then
+        vecho "NFSv4 server entry check: failed to convert $target_ip to hex; skipping guard."
+        return 1
+    fi
+
+    vecho "NFSv4 server entry check: looking for $target_ip:$target_port."
+
+    local line
+    while IFS= read -r line; do
```
sprasad-microsoft: Is this deliberate to unset IFS?

src/aznfswatchdogv4
```diff
@@ -10,6 +10,82 @@
...
+        if [ "${server_field,,}" != "$target_ip_hex" ]; then
+            continue
+        fi
+
+        if [ "${port_field,,}" = "$target_port_hex" ]; then
+            vecho "NFSv4 server entry check: matched entry (server=$server_field, port=$port_field)."
+            return 0
+        fi
+    done < "$NFSFS_SERVERS_FILE"
+    return 1
```
sprasad-microsoft: I see the return value is always 1 for this function. What's the purpose of the return value then?

sprasad-microsoft (approved):

src/aznfswatchdogv4
bharathsm-ms: if [ "${port_field,,}" = "$target_port_hex" ]; then
            vecho "NFSv4 server entry check: matched entry (server=$server_field, port=$port_field)."
            return 0
        fi
    done < "$NFSFS_SERVERS_FILE"
    return 1
    
    
Function returns 0 in case of matched entry otherwise 1

src/aznfswatchdogv4
```diff
@@ -10,6 +10,82 @@
...
+{
+    local ip=$1
+    local IFS=.
+    local o1 o2 o3 o4
+    read -r o1 o2 o3 o4 <<< "$ip"
+    if [ -z "$o1" -o -z "$o2" -o -z "$o3" -o -z "$o4" ]; then
+        return 1
+    fi
+    printf "%02x%02x%02x%02x" "$o1" "$o2" "$o3" "$o4"
+}
```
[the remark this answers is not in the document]
bharathsm-ms: This is not a problem for us because we always keep $LOCALHOST as 127.0.0.1. So, IP address we pass here will remain fixed or what we hardcoded in script as LOCALHOST.

src/aznfswatchdogv4
```diff
@@ -10,6 +10,82 @@
...
+        vecho "NFSv4 server entry check: missing target ip/port (ip='$target_ip', port='$target_port')."
+        return 1
+    fi
+
+    if [ ! -r "$NFSFS_SERVERS_FILE" ]; then
+        vecho "NFSv4 server entry check: $NFSFS_SERVERS_FILE not readable; skipping guard."
+        return 1
+    fi
+
+    target_port_hex=$(printf "%x" "$target_port")
```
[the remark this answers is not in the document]
bharathsm-ms: Thanks, Checked. /proc/fs/nfsfs/servers add the spaces as padding but not 0's. Since we trim spaces from the output before comparing, we are good.

src/aznfswatchdogv4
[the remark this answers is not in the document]
bharathsm-ms: @sprasad-microsoft Updated script to handle this comment. Please review

Issue #278, opened 2025-11-02
aznfs mount does not work on Azure China cloud due to cert rejected issue


andyzhangx:
mount.aznfs version: 0.3.15

I tried to use mount aznfs manually on the following Ubuntu node on Azure China  with EncryptionInTransit enabled, but it failed with following error due to cert rejected issue

# uname -a
Linux aks-sys-13702987-vmss000000 5.15.0-1096-azure #105-Ubuntu SMP Fri Aug 29 15:44:42 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux

mkdir /tmp/test
mount -v -t aznfs -o vers=4,minorversion=1,sec=sys f120618d2aa4140039707e1.file.core.chinacloudapi.cn:/f120618d2aa4140039707e1/pvcn-e1271d79-de32-44c5-a904-475c2ce1393f /tmp/test

Nov  2 03:42:18 aks-user-13702987-vmss000000 stunnel: LOG5[27186]: Connection reset: 0 byte(s) sent to TLS, 0 byte(s) sent to socket
Nov  2 03:42:18 aks-user-13702987-vmss000000 stunnel: LOG5[27187]: Service [40.73.81.200] accepted connection from 127.0.0.1:41154
Nov  2 03:42:18 aks-user-13702987-vmss000000 stunnel: LOG6[27187]: s_connect: connecting 40.73.81.200:2049
Nov  2 03:42:18 aks-user-13702987-vmss000000 stunnel: LOG5[27187]: s_connect: connected 40.73.81.200:2049
Nov  2 03:42:18 aks-user-13702987-vmss000000 stunnel: LOG5[27187]: Service [40.73.81.200] connected remote server from 10.224.0.5:42486
Nov  2 03:42:18 aks-user-13702987-vmss000000 stunnel: LOG6[27187]: SNI: sending servername: 40.73.81.200
Nov  2 03:42:18 aks-user-13702987-vmss000000 stunnel: LOG6[27187]: Peer certificate required
Nov  2 03:42:18 aks-user-13702987-vmss000000 stunnel: LOG4[27187]: CERT: Pre-verification error: unable to get local issuer certificate
Nov  2 03:42:18 aks-user-13702987-vmss000000 stunnel: LOG4[27187]: Rejected by CERT at depth=1: C=US, O=DigiCert Inc, CN=DigiCert Basic RSA CN CA G2
[...]

PR #294, opened 2026-02-09
Implement dynamic root CA detection for NFSv4 mount script

weijocelyn-ms:
In other clouds, DigiCert Root CA might not be available or may not be the correct root CA certificate to use to hit blob storage. This PR uses the storageaccount_ip to query the storage account's TLS certificate chain and determine the correct root CA, but falls back on DigiCert Root CA if this fails.

Related: https://github.com/Azure/AZNFS-mount/issues/278

### DIFF
--- a/src/nfsv4mountscript.sh
+++ b/src/nfsv4mountscript.sh
@@ -3,6 +3,12 @@
 #
 # NfSv4 logic for mount helper
 #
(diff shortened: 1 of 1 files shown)


andyzhangx:
I have verified that aznfs mount on Azure China East 2 region works now, I think we do not need this PR now

PR #298, opened 2026-03-03
Read Optimiziation for NFSv4

msftdan:
refactor read optimization to common.sh and add callers to nfsv3 and nfsv4.


bharathsm-ms (approved):
The below doc recommends setting to readhead size to ~15MB. 
https://learn.microsoft.com/en-us/azure/storage/files/nfs-performance .
"kernel 5.4 changed the NFS readahead from 15*rsize to fixed value of 128KB. So before 5.4 kernel for rsize=1M, NFS readahead was 15MB now it’s 128K irrespective of rsize."

Considering older kernel has readhead size of 15MB, change look good. But it will be a good experiment to compare across different workloads.

PR #303, opened 2026-05-01
[NFSv4] Support ZRS Zone failover 

danwong:


danwong:
> @danwong please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@microsoft-github-policy-service agree [company="Microsoft}"]

danwong:
> @microsoft-github-policy-service agree company="your company"

@microsoft-github-policy-service agree company=Microsoft

src/nfsv4mountscript.sh
```diff
@@ -43,6 +43,48 @@ CONNECT_PORT=2049
 # https://linux.die.net/man/5/nfs
 MOUNT_TIMEOUT_IN_SECONDS=180
 
+#
+# Variables for non-TLS proxy IP allocation (reuses NFSv3 IP allocation infrastructure).
+#
+DEFAULT_AZNFS_IP_PREFIXES="10.161 192.168 172.16"
+IP_PREFIXES="${AZNFS_IP_PREFIXES:-${DEFAULT_AZNFS_IP_PREFIXES}}"
```
sroghanchi: Since these are shared between v3 and v4 consider moving them to common.sh.

src/nfsv4mountscript.sh
```diff
@@ -591,7 +633,7 @@ tls_nfsv4_files_share_mount()
         # Mounted: mount command is executed successfully. If the mount is unmounted, watchdog can remove this entry.
         # Failed: mount command failed. Watchdog can remove this entry.
 
-        local mountmap_entry="$storageaccount_ip;$stunnel_conf_file;$stunnel_log_file;$stunnel_pid_file;$checksumHash;waiting;$mount_timeout"
+        local mountmap_entry="$nfs_host;$storageaccount_ip;$stunnel_conf_file;$stunnel_log_file;$stunnel_pid_file;$checksumHash;waiting;$mount_timeout"
```
sroghanchi, line 632: We need to make sure updating mountmap formatting doesn't break any existing functionality. For example, what if a customer already has AZNFS and some shares mounted with EiT and then updates AZNFS to this version.

src/nfsv4mountscript.sh
danwong: Done

src/aznfswatchdogv4
```diff
@@ -10,8 +10,264 @@
...
+# How often to probe the storage IP for reachability (seconds).
+HEALTH_CHECK_FREQUENCY=60
+
+#
+# Conntrack tracking hashes for NFSv4 non-TLS DNAT mounts.
+#
+declare -A cthash_synsent_v4
+declare -A cthash_unreplied_v4
+
+reconcile_conntrack_synsent_v4()
```
sroghanchi: Consider moving this and reconcile_conntrack_unreplied_v4 to common.sh to avoid duplicating code with nfsv3.

src/aznfswatchdogv4
```diff
@@ -10,8 +10,264 @@
...
+    fi
+
+    return $found_stuck
+}
+
+#
+# Force NFS client to send traffic through the updated DNAT rule
+# by issuing stat on the mount point.
+#
+ping_new_endpoint()
```
sroghanchi: Move to common.sh.

src/aznfswatchdogv4
```diff
@@ -227,6 +483,136 @@ cleanup_mount()
...
+    IFS=$'\n' notls_lines=$(cat $MOUNTMAPv4NOTLS)
+    flock -u $fd_notls
+    exec {fd_notls}<&-
+
+    if [ -z "$notls_lines" ]; then
+        return
+    fi
+
+    # Get current NFS mounts (done after reading mountmap for consistency).
+    local findmnt_notls=$(findmnt --raw --noheading -o SOURCE,TARGET -t nfs,nfs4 2>/dev/null)
```
sroghanchi, line 426: We could only filter for nfs4, do we need to include nfs as well?

sroghanchi, line 426: Exit if findmnt fails - similar to lines 652-658.

src/aznfswatchdogv4
```diff
@@ -227,6 +483,136 @@ cleanup_mount()
...
+        # Check 2: Verify DNAT rule still exists (safety net).
+        #
+        verify_iptable_entry "$l_ip" "$l_nfsip"
+
+        #
+        # Layer 1: Conntrack monitoring (passive, every 5s).
+        # Detects stuck SYN_SENT or UNREPLIED entries on port 2049.
+        #
+        local failure_detected=false
+        if reconcile_conntrack_v4 "$l_ip" "$l_nfsip"; then
```
sroghanchi, line 470: Is this something needs to be checked at every heath check (60 secs) or monitor check (5 secs)?

src/aznfswatchdogv4
```diff
@@ -10,8 +10,264 @@
...
+
+    # Update connect IP, pid path, log output path, and service section name in new conf.
+    sed -i "s/connect = ${old_ip}:2049/connect = ${new_ip}:2049/" $new_conf
+    sed -i "s#pid = ${l_pid}#pid = ${new_pid}#" $new_conf
+    sed -i "s#output = ${l_log}#output = ${new_log}#" $new_conf
+    sed -i "s/\[${old_ip}\]/[${new_ip}]/" $new_conf
+    chattr -f +i $new_conf
+
+    # Step 2: Kill old stunnel process and clean up old files.
+    if [ -f "$l_pid" ]; then
```
sroghanchi: Check if you can use the `cleanup_stunnel_files()` function here instead to eliminate duplicate code.

src/aznfswatchdogv4
```diff
@@ -395,7 +822,13 @@ if ! chattr -f +i $MOUNTMAPv4; then
     wecho "chattr does not work for ${MOUNTMAPv4}!"
 fi
 
+if ! chattr -f +i $MOUNTMAPv4NOTLS; then
+    wecho "chattr does not work for ${MOUNTMAPv4NOTLS}!"
+fi
+
 while :; do
     sleep $MONITOR_INTERVAL_SECS
+    epoch_now=$(date +%s)
```
sroghanchi, line 752: Remove the duplicate command on line 623.

sroghanchi (commented):
Make sure you test the package update:

- Install the latest version of the package from PMC
- mount some shares
- update the package to a version that has your changes
- try unmounting a share after update
- try mounting a new share which uses an existing connection (where two storage accounts share the same IP)
- try mounting a new share so you have new entries in the mountmap file with the new formatting - try unmounting a share with the old formatting to make sure it still works fine

src/aznfswatchdogv4
danwong: done.

src/nfsv4mountscript.sh
danwong: checked and it works properly as intended

src/aznfswatchdogv4
danwong: done.

danwong:
> Make sure you test the package update:
> 
> * Install the latest version of the package from PMC
> * mount some shares
> * update the package to a version that has your changes
> * try unmounting a share after update
> * try mounting a new share which uses an existing connection (where two storage accounts share the same IP)
> * try mounting a new share so you have new entries in the mountmap file with the new formatting - try unmounting a share with the old formatting to make sure it still works fine

done.

src/aznfswatchdogv4
danwong: Done.

src/aznfswatchdogv4
danwong: right now, it's on a 5s cycle, just like NFSv3. Then, the TCP Probe is done every 60 seconds.

src/aznfswatchdogv4
danwong: As discussed offline, it's possible that they would show up as nfs in findmnt. If we end up picking up further mounts, that don't have a matching mountmap file, then we are safe anyway.

danwong: discussed offline, it's possible that it would show up as nfs on certain distros. Anyway, watchdog would only iterate the mountmap file if it is listed anyway. I think it's safe. any objections?

src/aznfswatchdogv4
sroghanchi: I see, maybe add a comment clarifying that in this function cleanup is done every MONITOR_INTERVAL_SECS second and TCP probe every HEALTH_CHECK_FREQUENCY seconds.

lib/common.sh
```diff
@@ -343,6 +362,101 @@ is_private_ip()
     [[ $ip =~ ^192\.168\..* ]]
 }
 
+#
```
palashvij-msft, line 365: can we retain the NFSv3 side of comment here? as removed from aznfswatchdog, it's useful and more clear. You can add more for NFSv4 side additionally. Let's not shorten that one.

lib/common.sh
```diff
@@ -343,6 +362,101 @@ is_private_ip()
...
+    vecho "[$BASHPID] stat($target) #1 done"
+
+    sleep 35
+
+    vecho "[$BASHPID] stat($target) #2 start"
+    stat "$target"
+    vecho "[$BASHPID] stat($target) #2 done"
+}
+
+#
```
palashvij-msft, line 387: older comment would be better, mentioning why this required. it's easier to know older kernels have this issue.

src/aznfswatchdog
```diff
@@ -115,94 +115,6 @@ check_for_regional_account()
...
-    local seconds_remaining=$5
-
-    key="${l_ip}:${l_sport}:${l_dport}:${l_nfsip}"
-
-    # First time we are seeing this conntrack entry.
-    if [[ ! -v cthash_synsent[$key] ]]; then
-        cthash_synsent[$key]=$seconds_remaining
-        return
-    fi
-
```
palashvij-msft, line 141: let's have this comment back too

lib/common.sh
```diff
@@ -375,183 +489,444 @@ create_mountmap_file()
...
-            local mtime=$(stat -c%Y $MOUNTMAPv3)
+            local mtime=$(stat -c%Y $mountmap_file)
             if [ "$mtime" != "$ifmatch" ]; then
-                eecho "[$1] Refusing to remove from ${MOUNTMAPv3} as $mtime != $ifmatch!"
+                eecho "[$entry] Refusing to remove from ${mountmap_file} as $mtime != $ifmatch!"
                 return 1
             fi
         fi
 
-        # Delete iptable rule corresponding to the outgoing MOUNTMAPv3 entry.
```
palashvij-msft, line 444: let's add comment back

lib/common.sh
```diff
@@ -375,183 +489,444 @@ create_mountmap_file()
...
-        # We do this thing instead of inplace update by sed as that has a
-        # very bad side-effect of creating a new MOUNTMAPv3 file. This breaks
-        # any locking that we dependent on the old file.
-        #
-        out=$(sed "\%^${1}$%d" $MOUNTMAPv3)
+        chattr -f -i $mountmap_file
+        out=$(sed "\%^${entry}$%d" $mountmap_file)
         ret=$?
         if [ $ret -eq 0 ]; then
-            #
```
palashvij-msft, line 462: same, revert this comment

lib/common.sh
```diff
@@ -22,6 +22,19 @@ MOUNTMAPv3="${OPTDIRDATA}/mountmap"
...
+# This stores the map of hostname, local proxy IP and storage endpoint IP for NFSv4 non-TLS mounts.
+# Format: hostname localip storageip (same as MOUNTMAPv3)
+#
+MOUNTMAPv4NOTLS="${OPTDIRDATA}/mountmapv4notls"
+
+#
+# Default order in which we try the network prefixes for a free local IP to use.
+# This can be overridden using AZNFS_IP_PREFIXES environment variable.
+#
+DEFAULT_AZNFS_IP_PREFIXES="10.161 192.168 172.16"
```
palashvij-msft, line 35: Do we want to keep these ranges in common as well? I believe it might be more effective to have separate prefix ranges for v3 and v4. Is it essential to differentiate them? What do you think? V3 and V4- they now compete for the same local proxy IP pool.

lib/common.sh
```diff
@@ -375,183 +489,444 @@ create_mountmap_file()
...
+{
+    if [ "$AZNFS_PING_LOCAL_IP_BEFORE_USE" != "1" ]; then
+        return 1
+    fi
+
+    local ip=$1
+    ping -4 -W3 -c1 $ip > /dev/null 2>&1
+}
+
+#
```
palashvij-msft, line 801: comment can be kept same

lib/common.sh
```diff
@@ -375,183 +489,444 @@ create_mountmap_file()
...
+{
+    local ip=$1
+    local octet="[0-9]{1,3}"
+    local octetdot="${octet}\."
+
+    if ! is_valid_ipv4_prefix $ip; then
+        echo 0
+        return
+    fi
+
```
palashvij-msft, line 818: #
    # Check if the IP prefix belongs to the private IP range (10.0.0.0/8,
    # 172.16.0.0/12, or 192.168.0.0/16), i.e., will the user provided prefix
    # result in a private IP address.
    #

lib/common.sh
```diff
@@ -375,183 +489,444 @@ create_mountmap_file()
...
+    local local_ip=""
+    local optimize_get_free_local_ip=false
+    local used_local_ips_with_same_prefix=$(cat $MOUNTMAPv3 $MOUNTMAPv4NOTLS 2>/dev/null | awk '{print $2}' | grep "^${initial_ip_prefix}\." | sort -t . -k 1,1n -k 2,2n -k 3,3n -k 4,4n)
+    local iptable_entries=$(iptables-save -t nat)
+
+    _3rdoctet=100
+    ip_prefix=$initial_ip_prefix
+
```
palashvij-msft, line 867: #
    # Optimize the process to get free local IP by starting the loop to choose
    # 3rd and 4th octet from the number which was used last and still exist in
    # MOUNTMAPv3 instead of starting it from 100.
    #

lib/common.sh
danwong: yeah, I did a search of comments that I refactored with github copilot to see if there were any spots I missed. I'll run it again afterwards. But I think I restored most of them now.

lib/common.sh
danwong: reverted back to the original.

lib/common.sh
danwong: done and changed mountmapv3 to mountmap files.

lib/common.sh
danwong: I don't think it's necessary to split these unless you have a strong feeling otherwise. 
search_free_local_ip_with_prefix still allocates them properly and is not essential for correctness. Splitting could lead to more complexity.

src/aznfswatchdog
danwong: done

palashvij-msft (approved):

PR #304, opened 2026-05-07
Add SLES16 package support

sroghanchi:


lib/common.sh
```diff
@@ -163,7 +163,7 @@ systemd_is_init()
 ensure_aznfswatchdog()
 {
     local process_name="$1"
-    pidof -x "$process_name" > /dev/null 2>&1
+    pgrep -x "$process_name" > /dev/null 2>&1
```
palashvij-msft, line 166: can we check something like
command -v pgrep? and fail gracefully if needed? what do u think? we can discuss, if not pgrep then pidof can be tried?

palashvij-msft (approved):

lib/common.sh
sroghanchi: Using pgrep should be safe here as we already have use it:

https://github.com/Azure/AZNFS-mount/blob/a0257f0a4d36f98dc01332e30fa01afb44c991d3/packaging/aznfs/RPM/aznfs.spec#L158
 
https://github.com/Azure/AZNFS-mount/blob/a0257f0a4d36f98dc01332e30fa01afb44c991d3/packaging/aznfs/RPM/aznfs.spec#L158
 
https://github.com/Azure/AZNFS-mount/blob/a0257f0a4d36f98dc01332e30fa01afb44c991d3/scripts/aznfs_install.sh#L223

PR #313, opened 2026-06-08
Turbo ACL Bug Fix

DevanshG1:
Since the turbo client runs as a FUSE demon, it has a uid=0. When root squash is enabled, even though sattr3 has the correct uid/gid, it is not honored and the uid used is 0. To fix this i have called the setuid and setguid functions before each RPC dispatch

### DIFF
--- a/turbonfs/inc/rpc_task.h
+++ b/turbonfs/inc/rpc_task.h
@@ -2120,6 +2120,29 @@ public:
         return nfs_get_rpc_context(get_nfs_context());
     }
 
+    /**
+     * Set RPC AUTH_UNIX credentials to match the FUSE caller's uid/gid.
+     *
+     * The FUSE daemon runs as root, so libnfs defaults to uid=0/gid=0
+     * in the RPC AUTH_UNIX header. With root_squash enabled on the
+     * server, this causes ALL operations to be squashed to nobody.
+     *
+     * This must be called before the rpc_nfs3_*_task() call so that
+     * rpc_allocate_pdu() picks up the correct AUTH credentials.
+     * Thread safety: nfs_set_uid/gid internally calls rpc_set_uid_gid
+     * which replaces rpc->auth. The rpc_mutex inside libnfs protects
+     * the auth update and PDU allocation atomically.
+     */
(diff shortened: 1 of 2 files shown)


turbonfs/inc/rpc_task.h
```diff
@@ -2120,6 +2120,43 @@ struct rpc_task
...
+
+        if (req == nullptr) {
+            return;
+        }
+
+        const fuse_ctx *ctx = fuse_req_ctx(req);
+        if (ctx) {
+            struct nfs_context *nfs = get_nfs_context();
+            nfs_set_uid(nfs, ctx->uid);
+            nfs_set_gid(nfs, ctx->gid);
```
linuxsmiths, line 2156: you may also want to set the supplementary groups for complete permission sanity, o/w some permissions granted via supplementary groups may not work.
Do check what supplementary groups are being sent in the RPC.

turbonfs/src/rpc_task.cpp
```diff
@@ -1568,6 +1568,7 @@ void rpc_task::issue_write_rpc()
 
     do {
         rpc_retry = false;
+        set_caller_credentials();
```
linuxsmiths, line 1571: the change to set the uid/gid in the RPC, from the fuse ctx is correct, just make sure you do it for all NFS command and none get missed out.

linuxsmiths (commented):
LG o/w

linuxsmiths (commented):
I realize that I had this unsent review lying..

turbonfs/inc/rpc_task.h
DevanshG1: Done

turbonfs/src/rpc_task.cpp
DevanshG1: Done

turbonfs/inc/rpc_task.h
linuxsmiths: I don't see supplementary groups being set.

linuxsmiths (approved):
LG

PR #320, opened 2026-07-20
turbonfs: remove GnuTLS dependency, non-TLS only

pragyagandhi:
## Summary

Remove the **GnuTLS** dependency from the turbonfs build so builds are not blocked on distros
where gnutls is not an allowed package (e.g. Azure Linux). The only supported transport is
non-TLS (`xprtsec=none`), and **AZAUTH (AzAuthNone) is always sent as the first RPC** over that
non-TLS connection.

This is the turbonfs/packaging side. The corresponding libnfs changes are in
linuxsmiths/libnfs#39.

## Changes

- **turbonfs/CMakeLists.txt**
  - Drop `find_package(GnuTLS)` and the static gnutls linking
    (`libgnutls`/`hogweed`/`nettle`/`tasn1`/`idn2`/`gmp`/`p11-kit`/`unistring`) — the binary no
    longer links gnutls.
  - Default `ENABLE_INSECURE_AUTH_FOR_DEVTEST` to **ON**.
- **turbonfs/inc/util.h**: `is_valid_xprtsec()` accepts only `none`.
- **turbonfs/build.sh, package.sh, generate_package.sh**: set
  `INSECURE_AUTH_FOR_DEVTEST=ON` for all build types (not just Debug).

## Why the flag is required for all builds

libnfs now has TLS removed, so AZAUTH is always sent over the non-TLS connection. That path is
gated by `ENABLE_INSECURE_AUTH_FOR_DEVTEST`, and libnfs bails out (static gate) if it is
disabled. Hence turbonfs must enable it for every build type — previously it was Debug-only.

## Runtime flow

```mermaid
sequenceDiagram
    participant C as aznfsclient (turbonfs)
    participant L as libnfs
    participant S as Blob NFS server (:2048)
    C->>L: nfs_set_auth_context(authtype=AzAuthNone) -> use_azauth=TRUE
    C->>L: nfs_mount()
    L->>S: TCP connect (non-TLS)
    L->>S: AZAUTH RPC (AzAuthNone)  (always first, over non-TLS)
    alt server has AzAuth enabled
        S-->>L: AZAUTH OK
        L->>S: MOUNT / NFS ops
        S-->>C: mounted (or MNT3ERR_NOENT if container missing)
    else server without AzAuth
        S-->>L: no response / reject
        L-->>C: "AzAuth not enabled/setup on server" diagnostic
    end
```

## Testing

- Full turbonfs build succeeds; `aznfsclient` links **no** gnutls
  (`ldd` shows only libuuid/libm/libc).
- AzAuth-enabled account: AzAuthNone accepted, proceeds to MOUNT (server responds).
- Non-AzAuth account: AZAUTH fails fast with a clear diagnostic.

> Depends on linuxsmiths/libnfs#39 (submodule pointer bump to be done once that merges).

### DIFF
--- a/turbonfs/sample-turbo-config.yaml
+++ b/turbonfs/sample-turbo-config.yaml
@@ -12,7 +12,7 @@
 # account name, e.g., "myaccount-secondary".
 #
 account: nfsv3acc1bl2prdev13a
-container: testcont
+container: nfsv3test
 
 #cloud_suffix: blob.core.windows.net
 cloud_suffix: blob.preprod.core.windows.net
@@ -62,7 +62,7 @@ sys.nodrc.create_exist_as_success: true
 #
 
 # valid range: [1, 256]
-nconnect: 96
+nconnect: 1
 
 # valid range: [100, 6000]
 timeo: 600
@@ -203,9 +203,10 @@ oom_kill_disable: true
 auth: false
 
 #
-# Option controlling transport security, takes following values:
+# Option controlling transport security. Only "none" is supported.
+# RPC-with-TLS support (and its GnuTLS dependency) has been removed, so the
+# transport is always non-TLS.
 # none: do not use TLS encryption.
-# tls: use TLS encryption, as defined in RPC-with-TLS RFC 9289.
 #
 xprtsec: none
 
--- a/turbonfs/sample-turbo-config.yaml
+++ b/turbonfs/sample-turbo-config.yaml
@@ -11,11 +11,11 @@
 # For accessing the secondary endpoint (RA-GRS), append "-secondary" to the
 # account name, e.g., "myaccount-secondary".
 #
-account: sjc22prdste06hnfsv3acc1
-container: nfsv3test
+account: nfsv3acc1bl2prdev13a
+container: testcont
 
 #cloud_suffix: blob.core.windows.net
-#cloud_suffix: blob.preprod.core.windows.net
+cloud_suffix: blob.preprod.core.windows.net
 #port: 2048
 
 #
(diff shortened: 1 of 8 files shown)


turbonfs/sample-turbo-config.yaml
```diff
@@ -11,11 +11,11 @@
...
 # account name, e.g., "myaccount-secondary".
 #
-account: sjc22prdste06hnfsv3acc1
-container: nfsv3test
+account: nfsv3acc1bl2prdev13a
+container: testcont
 
 #cloud_suffix: blob.core.windows.net
-#cloud_suffix: blob.preprod.core.windows.net
+cloud_suffix: blob.preprod.core.windows.net
```
palashvij-msft, line 18: let's comment this out as earlier

palashvij-msft (approved):

turbonfs/sample-turbo-config.yaml
pragyagandhi: This is a working preprod tenant and account, hence have added, older one fails anyway

palashvij-msft: Yes, agreed but we put this sample turbo config as part of release also.
and if no config is provided, this is default used. 
and we don't want default cloud_suffix: blob.preprod.core.windows.net to be picked. usually our server.cpp or auto checks for all cloud suffixes.

correct me, might be missing a thing?

palashvij-msft: I checked, release/3.0 doesn't have these extra params. should be good. config in main and release/3.0 are seperate

turbonfs/sample-turbo-config.yaml
palashvij-msft, line 1: L208, Documentation still advertises TLS

linuxsmiths (approved):

PR #331, opened 2026-09-14
Adding AZNFS-mount support on Azure Linux 3.0

psachdeva-ms:
This pull request adds support for Azure Linux 3.0 to the build, packaging, and release pipeline for AZNFS. The changes ensure that Azure Linux RPM packages are correctly built, signed, published, and documented. Additionally, the installation scripts and packaging logic are updated to use the appropriate package manager (`tdnf`) for Azure Linux.

**Azure Linux 3.0 Support**

* Added Azure Linux 3.0 (both x86_64 and ARM64) to the list of supported distributions in `README.md` and included the appropriate entries in `packages.csv` for publishing. 
* Updated the build pipeline (`.github/aznfs-build.yaml`) to:
  - Add ESRP code signing steps for Azure Linux 3.0 RPMs (x86_64 and ARM64).
  - Upload signed Azure Linux RPMs to the package management system.
  - Remove temporary cleanup of Azure Linux RPM artifact directories to ensure artifacts are available for publishing.

**Packaging and Installation Logic**

* Modified `generate_package.sh` and `package.sh` to use `tdnf` as the install command for Azure Linux RPMs, while keeping `yum` for other RPM-based distros.
* Updated `scripts/aznfs_install.sh` to prefer `tdnf` (then `dnf`, then `yum`) for package installation, improving compatibility with Azure Linux.


scripts/aznfs_install.sh
rajasi3010: Both [ $? -eq 1 ] checks are dead code for tdnf. tdnf returns 0 even when updates are available (the 100 return is gated behind dnf_check_update_compat in tdnf.conf, off by default), and its error codes are 900/1000–2600 — it never returns 1. So a repo or network failure on Azure Linux won't trip the exit 1 guard; $yum list available then returns nothing and auto-update becomes a silent no-op.

scripts/aznfs_install.sh
```diff
@@ -102,7 +102,11 @@ vecho()
 use_dnf_or_yum() 
 {
     yum="yum"
-    if command -v dnf &> /dev/null; then
+    # Azure Linux ships only tdnf by default; prefer it, then dnf, then yum.
+    if command -v tdnf &> /dev/null; then
```
rajasi3010: Any RHEL/Rocky/CentOS/Oracle machine that happens to have tdnf installed silently switches package manager mid-upgrade. Should we use "$distro_id" == "azurelinux" here?

.github/aznfs-build.yaml
```diff
@@ -434,6 +433,60 @@ stages:
...
+              Pattern: '*.rpm'
+              SessionTimeout: 90
+              ServiceEndpointUrl: 'https://api.esrp.microsoft.com/api/v2'
+              MaxConcurrency: 25
+              signConfigType: inlineSignParams
+              VerboseLogin: true
+              inlineOperation: |
+                [
+                  {
+                    "KeyCode": "CP-459159-Pgp",
```
rajasi3010: CP-459159-Pgp is the right keycode for Azure Linux 3.0 , but it's the first time this pipeline uses it — unlike the other keycodes, it isn't onboarded yet. Two prerequisites before merge:

ESRP sign permission for CP-459159-Pgp on the AZNFS app registration (requested via ESRP, check PMC docs). Without it the task fails .
PMC publish access to azurelinux-3.0-prod-ms-oss-{x86_64,aarch64}-yum. So you will see their yaml files in PMC repo does not have group shred and aznfs can only access shared/test repo

